PR #23, opened 2020-03-18
Update description field to take any prop

maxlew:
This update allows the description prop in ImageWithHeaderAndContent to take any value.

Note: this PR causes the custom styling that was previously on the `<Paragraph>` element to be lost.


beau-charman:
Thank you for picking this up Max

beau-charman:
In regards to the <p /> styling, you can use ParagraphSet https://github.com/aesop/aesop-gel/blob/develop/src/components/Paragraph/Paragraph.js#L24 to wrap any child <p /> tags in GEL styling. 

Also merge conflict.

maxlew:
@beau-charman good point about the Paragraph. I'll use that as a wrapper to control styling. Happy to merge this now if you are.

beau-charman:
👍

PR #40, opened 2020-05-08
[Feature][AES-828] Audio Player Compoment

beau-charman:
This PR introduces the Audio Player Component 

GEL Storybook: https://aesop.github.io/aesop-gel/?path=/story/components-audio--base-component

Addresses: https://aesoponline.atlassian.net/browse/AES-828


src/components/Audio/Audio.js
```diff
@@ -0,0 +1,272 @@
...
+          <SeekForwardButton />
+          <PausePlayButton />
+          <SeekBackwardButton />
+        </div>
+        <div className={styles.download}>
+          <Hyperlink
+            className={styles.downloadButton}
+            hasTargetInNewWindow={true}
+            isDownload={true}
+            style="External No Icon Link"
```
maxlew, line 216: This should probably be a constant right?

src/components/Audio/Audio.utils.js
```diff
@@ -0,0 +1,64 @@
...
+const drawLineSegment = (ctx, x, height) => {
+  ctx.lineWidth = 2;
+  ctx.strokeStyle = '#fff';
+  ctx.beginPath();
+  ctx.moveTo(x, 0);
+  ctx.lineTo(x, -height);
+  ctx.stroke();
+};
+
+drawAudio('https://s3-us-west-2.amazonaws.com/s.cdpn.io/3/shoptalk-clip.mp3');
```
maxlew, line 64: Left over from testing?

src/components/Audio/Audio.js
beau-charman: Yes it should

src/components/Audio/Audio.utils.js
beau-charman: Left over from manually creating the wave graph, now I am using the wavesurfer package. I kept this file incase I needed to revert, but am happy to remove and retrieve from git history if needed.

beau-charman:
In regards to the mp3 @maxlew , I already ran compressed it at 32kbps. I might run it through a few more times to squeeze more out. You're right, it doesn't need to sound any good

PR #47, opened 2020-05-20
[fix] typo on podium padding rules

maxlew:
Found a small typo when generating the podium padding class names.


beau-charman:
Thank you for the pick up @maxlew 😓

PR #50, opened 2020-05-28
[Feature][AES-1160] Map Component

beau-charman:
This PR introduces the Map component. It utilises the `google-maps` and `@google/markerclustererplus` packages.


src/components/GoogleMap/GoogleMap.js
```diff
@@ -0,0 +1,359 @@
...
+          new MarkerClusterer(mapRef.current, markers, {
+            imageExtension: 'png',
+            imagePath: GOOGLE_MAPS.CLUSTER_IMAGE_PATH,
+          }),
+      );
+    }
+    return function cleanup() {
+      clearMapClusters();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
maxlew, line 131: Guessing it's the `googleMap` variable that's being deliberately ignored from the useEffect deps here. Is there a side effect to adding it?

src/customHooks/useScript.js
```diff
@@ -1,25 +1,80 @@
-import { useEffect } from 'react';
+import { useEffect, useState, useRef } from 'react';
+
+const isBrowser =
+  typeof window !== 'undefined' && typeof window.document !== 'undefined';
+
+const useScript = ({
```
maxlew, line 6: I like this hook! Loading external scripts inside of react seems like a weird pattern. But surprisingly quite a lot of 3rd party tools opt for the script tag over an npm module, gives them better version control I guess.

src/components/GoogleMap/GoogleMap.js
beau-charman: `googleMap` refers to the google map library (loads script tags and styles to the html), and I am going to great limits to prevent it from being reloaded, hence not needing to listen for it

src/customHooks/useScript.js
beau-charman: yeah! and the hook cleans up the process hopefully. and based on my comment above, i might even try and load googleMaps through useScript, rather than `google-maps` loader. but only if i get more issues.

src/components/GoogleMap/GoogleMap.js
maxlew: Ah makes sense

beau-charman: hence the higher level component context, even if there are multiple maps on a page, there should only be one googleMaps library 'Provider'

PR #51, opened 2020-06-15
[Feature] Podium Update

beau-charman:
This PR adds a `backgroundColor` prop to the Podium component. 

It also updates the carousel component to better handle the `light` theme prop.


beau-charman:
Yeah, was thinking of doing that separately but could just smash it out in one ticket - will just have to coordinate with the MediaWithContent component

PR #60, opened 2020-07-01
[Update] Store Detail Header Links 

beau-charman:
This PR adds Hyperlink components to the Store Detail Header component to link the Location and Email attribute to their associated functionality.


beau-charman:
> are you able to add the `StoreDetailHeader` export to `src/components/index.js`

Good call, added

PR #64, opened 2020-07-09
[Feature][AES-1277] Refactor contexts, hooks and implementation of AddToCart, products and variants

beau-charman:
This is a refactor PR to prepare for https://aesoponline.atlassian.net/browse/AES-1277

- Refactor contexts, hooks and implementation of AddToCart, products and variants


beau-charman:
`ProductDeatailHeader` and it's children components are for consumption in web-ui. The ProductHeader composition is 90% for demo purposes, and even though I am sure sure of how the ProductDetailsHeader will be implemented into web-ui, its also an option as a composition component. In that even we will update the props (particularly the underscore prefixed ones) to suit the needs of web-ui. Cheers @maxlew

PR #79, opened 2020-07-24
[Feature][AESS-1496] Two Column Image Carousel

beau-charman:
This PR makes adjustments to the TwoColumnLayout and Podium to accommodate an Image Carousel component correctly 

https://aesoponline.atlassian.net/browse/AES-1496


src/components/Podium/Podium.js
```diff
@@ -11,7 +11,13 @@ const Podium = forwardRef(function PodiumRef(
     className,
     horizontalPadding,
     isActive,
+    isActiveOnMount,
+    isHorizontalFlushOnSmall,
+    isHorizontalFlushOnMedium,
+    isHorizontalFlushOnLarge,
```
zack76, line 17: Just wondering if these should be alphabetically sorted as well?

PR #100, opened 2020-08-06
[Update/AES-1350]: update hub article url for testing

zack76:
add the query `'?contentful=true'` to uri passed in content hub article for development


src/components/ReadMore/__snapshots__/ReadMore.spec.js.snap
```diff
@@ -178,7 +178,7 @@ exports[`<ReadMore /> renders base component correctly 1`] = `
       <a
         className="base left dark title"
         data-test-ref="test_3_TITLE"
-        href="https://www.aesop.com/au/r/careers"
+        href="https://www.aesop.com/au/r/careers?contentful=true"
```
beau-charman, line 181: might be good at a later date to create a feature flag for this sort of thing

PR #116, opened 2020-08-14
[Feature][AES-1748] Read Page typography updates

beau-charman:
This PR address some typography updates in https://aesoponline.atlassian.net/browse/AES-1748

It also creates a Theme Context to aid in passing shared themes through descendant Components


src/components/Heading/Heading.module.css
```diff
@@ -38,61 +38,49 @@
...
+    font-size: rem(26px);
+
+    &.serifFont {
+      font-size: rem(30px);
+    }
   }
 
   @media (--viewport-lg) {
-    font-size: rem(30px);
+    font-size: rem(24px);
```
zack76, line 80: i have a similar ticket which indicates the font size for Heading 'Suisse' is 30px for desktop view, 26px for tablet and 24px for mobile view. But what Max left in the comments are not same and do not match the equivalent module. I think we shoud double check with UX team

beau-charman: What ticket in particular?

zack76: AES-1346. in the invision design

beau-charman: Thank you for catching this one Zack

PR #124, opened 2020-08-18
Feature/aes 1180 product commerce slice padding

MkMan:
Adding padding top and bottom to the product commerce slice.

product commerce


src/components/ProductCommerce/ProductCommerce.module.css
```diff
@@ -13,6 +13,62 @@
   }
 }
 
+.nonePaddingBottom {
```
MkMan, line 16: The content model specified the same values as Podium component so I copy/pasted those classes from it to maintain a consistent look.

Is there a way to make those classes common to avoid duplicate code?

src/components/ProductCommerce/ProductCommerce.js
```diff
@@ -45,7 +47,12 @@ const ProductCommerce = ({
   }
 
   const variantRadioOptions = getVariantRadioOptions(variants);
-  const classSet = cx(styles.base, className);
+  const classSet = cx(
+    styles.base,
+    styles[`${paddingBottom.toLowerCase()}PaddingBottom`],
```
MkMan, line 52: Currently,  the content model has the options as `None` | `Small` etc... I reached out to the PO to make them lower case but until then I thought it's best to convert them manually to not block the story.

MkMan:
Are the commit messages and branch name ok to trigger a version bump?

src/components/ProductCommerce/ProductCommerce.module.css
zack76: i am thinking using Podium as a wapper so you don't have to pass padding to this component

MkMan: Yes, I think you're right Zack. Thanks for pointing that out. I'll update the PR.

MkMan:
Closing as these changes aren't needed. The padding will be set by the wrapping `<Podium>` component.

PR #144, opened 2020-08-27
[Feature][AES-1744] Hyperlink and Button Style Updates

beau-charman:
This PR addresses the button style updates in Max's comment on https://aesoponline.atlassian.net/browse/AES-1744

https://www.figma.com/file/pXcRya6xAK8hxUGkd22Xpr/Body-Copy-CTA's?node-id=0%3A1

### DIFF
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -17,23 +17,33 @@
   > * {
     pointer-events: none;
   }
+
+  &,
+  & * {
+    box-sizing: border-box;
+  }
 }
 
 .blockStyle {
   display: inline-block;
   min-width: 220px;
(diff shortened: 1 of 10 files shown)


src/components/Button/Button.module.css
```diff
@@ -17,23 +17,33 @@
   > * {
     pointer-events: none;
   }
+
+  &,
+  & * {
```
MkMan, line 22: What does `&,` do?

MkMan (approved):

src/components/Button/Button.module.css
MkMan: Awesome, thank you

PR #145, opened 2020-08-30
[Feature][AES-1783] Carousel Pagination Animation

beau-charman:
This PR adds animation to carousel pagination to address https://aesoponline.atlassian.net/browse/AES-1778

### DIFF
--- a/src/components/Carousel/components/Pagination/Pagination.js
+++ b/src/components/Carousel/components/Pagination/Pagination.js
@@ -4,18 +4,20 @@ import PropTypes from 'prop-types';
 import styles from './Pagination.module.css';
 
 const Pagination = ({ dots, hasFlushPagination, progressIndex, theme }) => {
-  const width = `${100 / dots.length}%`;
-  const positionLeft = `${(100 / dots.length) * progressIndex}%`;
+  if (!dots.length) return null;
+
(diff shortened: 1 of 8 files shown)


src/components/Carousel/components/Pagination/Pagination.js
```diff
@@ -1,40 +1,38 @@
...
+  const positionLeft = `${(100 / dots.length) * progressIndex}%`;
 
   return (
     <div className={styles.base}>
-      <ul
-        className={cx(styles.list, styles[theme], {
-          [styles.flush]: hasFlushPagination,
-        })}
+      <div
+        className={cx(
```
MkMan, line 13: nitpick: Do we have a preference for where we define classes? So far I have seen them declared prior to use in a constant which seems nicer because it leaves the jsx lean.

MkMan (approved):

src/components/Carousel/components/Pagination/Pagination.js
beau-charman: Depending on the complexity of the component, I set classSets if there are more than a few html elements, or multiple classSets, or a large amount of logic to generate the classSets. As this is a small component, with no need to scale, creating a new memory pointer is not really required, happy to change for consistency though

PR #146, opened 2020-08-31
Update webui from gel

MkMan:
### What
These changes will update the version of Aesop Gel in Aesop Web UI every time a new version (tag) of Gel is published.

### Why
Very often multiple PR on Aesop Web UI are updating the version of Gel as part of their changes. This can cause conflicts at the time of merge and may require rebasing to drop changes to `package.json` and `package-lock.json`.

### How
1. At the time of release the version number is stored
2. After release, the CI will checkout `develop` of Aesop Web UI and run an npm install to update the version of gel to the that was just published
3. The changes are staged, committed and pushed

#### Example:

I created a testing branch and reset Gel version on it (myself) to an old version. After running the modified GitHub action proposed in this PR the branch was updated

image


.github/workflows/push.yml
```diff
@@ -28,6 +28,27 @@ jobs:
...
+      - name: Checkout Web UI
+        uses: actions/checkout@v2
+        with:
+          repository: 'aesop/aesop-web-ui'
+          fetch-depth: 1
+          ref: 'develop'
+          token: ${{ secrets.X_GITHUB_TOKEN }}
+
+      - name: Update Gel in Web UI
+        run: npm install --save github:aesop/aesop-gel#$RELEASED_GEL_VERSION
```
MkMan, line 46: I tried running this without actually downloading the packages using the `--package-lock-only` flag but it complained about not having access to the Gel repo 😕

.github/workflows/push.yml
```diff
@@ -28,6 +28,27 @@ jobs:
...
 
       - name: Create Release
-        run: npx semantic-release
+        run: |
+          npx semantic-release
+          echo "::set-env name=RELEASED_GEL_VERSION::$(<version.txt)"
         env:
           GITHUB_TOKEN: ${{ secrets.X_GITHUB_TOKEN }}
+
+      - name: Checkout Web UI
```
MkMan, line 37: I initially tried doing the steps from this point onwards in a separate shell script file. I faced an issue when cloning Aesop Web UI due to SSH access. So instead I opted for using the `checkout` action provided by GitHub.

.releaserc
```diff
@@ -17,6 +17,9 @@
       "successComment": false,
       "releasedLabels": false,
     }],
+    ["@semantic-release/exec", {
+      "successCmd": "echo v${nextRelease.version} > version.txt"
```
MkMan, line 21: The only alternative I could find to doing this is using `git fetch` and `git describe` to get the tag. This is more robust as it gets the tag/version as part of the release process.

maxlew:
I really like this, but this there's a bunch of process things we need to sort out first. 

I feel like this action would be better suited on a master merge for this repo, the develop merges are supposed to be 'pre-release' versions, hence the `-develop.x` tags. You could create another action that fires on push but uses a tag filter instead of branch, something like below
```
on:
  push:
    tags: 
      - v* 
    tags-ignore:
      - *-develop.*
```
That'd also mean you had access to the version number easier as it would be your `GITHUB_REF` environment variable.

Also I think directly updating web-ui is a little extreme, ideally this would create a PR in any repo that uses aesop-gel notifying of an update, similar to how something like dependabot works, if this was a published NPM repo we could actually use dependabot.

MkMan:
> I really like this, but this there's a bunch of process things we need to sort out first.
> 
> I feel like this action would be better suited on a master merge for this repo, the develop merges are supposed to be 'pre-release' versions, hence the `-develop.x` tags. You could create another action that fires on push but uses a tag filter instead of branch, something like below
> 
> ```
> on:
>   push:
>     tags: 
>       - v* 
>     tags-ignore:
>       - *-develop.*
> ```
> 
> That'd also mean you had access to the version number easier as it would be your `GITHUB_REF` environment variable.
> 
> Also I think directly updating web-ui is a little extreme, ideally this would create a PR in any repo that uses aesop-gel notifying of an update, similar to how something like dependabot works, if this was a published NPM repo we could actually use dependabot.

Good points @maxlew 👍 

I guess I noticed this problem when myself and others were trying to update gel to `1.2.0-develop.xx` so I tailored the solution towards develop tags. From what I have noticed so far, we tend to reference develop tags instead of master releases (which are less often) so an update for master releases only won't solve the problem.

I like the idea of raising a PR, I'll look into it a bit further and move this PR to draft in the meantime.

maxlew:
Yeah we haven't had many master releases of gel so far. I think it's very easy to just use the develop tags and that's what has been happening, I feel like while this repo is under pretty rapid development it makes sense to keep that barrier as low as possible. 

Curious to know what people think of adding an extra step to merge aesop-gel branches into master in order for them to be available in web-ui. You could still use the 'develop' versions during development but we'd only want to use proper stable marked releases for develop/production

PR #147, opened 2020-08-31
Raise pr on tag push

MkMan:
### What
These changes will update the version of Aesop Gel in Aesop Web UI every time a new version (tag) of Gel is published. The update is done on a separate branch and PR is created to merge it to develop.

### Why
Very often multiple PR on Aesop Web UI are updating the version of Gel as part of their changes. This can cause conflicts at the time of merge and may require rebasing to drop changes to `package.json` and `package-lock.json`.

### How
1. The workflow runs for every tag push
2. It checks out Aesop Web UI
3. It runs a custom [GitHub action](https://github.com/aesop/aesop-github-action/tree/create-pr-action/create-pr)

#### Example:

[PR 1](https://github.com/aesop/aesop-web-ui/pull/3162): this was created on tag push. Then closed when another tag was pushed.
[PR 2](https://github.com/aesop/aesop-web-ui/pull/3163) This was created upon push of a new tag. It closed the above unmerged PR.

### DIFF
--- a/.github/workflows/update-webui.yml
+++ b/.github/workflows/update-webui.yml
@@ -29,10 +29,24 @@ jobs:
       - name: Update Gel
         run: npm install --save github:aesop/aesop-gel#$GEL_VERSION
 
+      - name: Commit and push update
+        run: |
+          git config user.name "Jenkins Aesop"
+          git config user.email ""
+          git add ./package*
+          git checkout -b $BRANCH_NAME
+          git commit -m "chore(update-gel): update Aesop Gel to ${GEL_VERSION}"
+          git push --set-upstream origin $BRANCH_NAME
+
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v3
-        with:
-            token: ${{ secrets.X_GITHUB_TOKEN }}
-            commit-message: Update Gel to ${{ env.GEL_VERSION }}
-            title: Update Gel to ${{ env.GEL_VERSION }}
-            branch: update-gel-to-${{ env.GEL_VERSION }}
+        run: |
+          curl \
+          -u "jenkins-aesop:${TOKEN}" \
+          -X POST \
+          -H "Accept: application/vnd.github.v3+json" \
+          https://api.github.com/repos/aesop/aesop-web-ui/pulls \
+          -d "{\"title\":\"Update Aesop Gel to ${GEL_VERSION}\",\"head\":\"${BRANCH_NAME}\",\"base\":\"develop\"}"
+        env:
+          TOKEN: ${{ secrets.X_GITHUB_TOKEN }}
+          VERSION: ${{ env.GEL_VERSION }}
+          BRANCH_NAME: ${{ env.BRANCH_NAME }}
--- a/.github/workflows/update-webui.yml
+++ b/.github/workflows/update-webui.yml
@@ -17,17 +17,22 @@ jobs:
           ref: 'develop'
           token: ${{ secrets.X_GITHUB_TOKEN }}
 
+      - name: Create variables
+        run: |
+          echo "::set-env name=GEL_VERSION::${GITHUB_REF#refs/*/}"
+          echo "::set-env name=BRANCH_NAME::update-Gel-to-${GEL_VERSION}"
+
       - name: npm install
-        run: npm install --save github:aesop/aesop-gel#$GITHUB_REF
+        run: npm install --save github:aesop/aesop-gel#$GEL_VERSION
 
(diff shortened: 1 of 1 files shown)


.github/workflows/update-webui.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Create variables
+        run: |
+          echo "::set-env name=GEL_VERSION::${GITHUB_REF#refs/*/}"
+          echo "::set-env name=BRANCH_NAME::update-Gel-to-${GITHUB_REF#refs/*/}"
+
+      - name: Update Gel
+        run: npm install --save github:aesop/aesop-gel#$GEL_VERSION
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
```
maxlew, line 33: All looks pretty good. The `${{ secrets.X_GITHUB_TOKEN }}` is pretty sensitive though so will need to do a bit of a review of https://github.com/peter-evans/create-pull-request it's probably worth us forking so that we aren't vulnerable to someone pushing some dodgy code there down the line.

MkMan: I have changed it to use Github API and not an external action :slightly_smiling_face:

beau-charman (approved):

.github/workflows/update-webui.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          repository: 'aesop/aesop-web-ui'
+          fetch-depth: 1
+          ref: 'develop'
+          token: ${{ secrets.X_GITHUB_TOKEN }}
+
+      - name: Create variables
+        run: |
+          echo "::set-env name=GEL_VERSION::${GITHUB_REF#refs/*/}"
+          echo "::set-env name=BRANCH_NAME::update-Gel-to-${GITHUB_REF#refs/*/}"
```
maxlew, line 27: Might be good to make this a cleaner branch name. Using dependabot as an example they use something like 
`dependabot/npm_and_yarn/graphql-tools/schema-6.2.0` so we could do something like
`dependencies/aesop-gel-${GEL_VERSION}`

.github/workflows/update-webui.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Update Gel
+        run: npm install --save github:aesop/aesop-gel#$GEL_VERSION
+
+      - name: Commit and push update
+        run: |
+          git config user.name "Jenkins Aesop"
+          git config user.email ""
+          git add ./package*
+          git checkout -b $BRANCH_NAME
+          git commit -m "chore(update-gel): update Aesop Gel to ${GEL_VERSION}"
```
maxlew, line 38: Again using dependabot as an example this should be something like 
`build(deps): bump aesop-gel from 6.0.9 to 6.2.0` having the from might be a hard to happy to just have the `build(deps):` part

.github/workflows/update-webui.yml
```diff
@@ -0,0 +1,52 @@
...
+          git push --set-upstream origin $BRANCH_NAME
+
+      - name: Create Pull Request
+        run: |
+          curl \
+          -u "jenkins-aesop:${TOKEN}" \
+          -X POST \
+          -H "Accept: application/vnd.github.v3+json" \
+          https://api.github.com/repos/aesop/aesop-web-ui/pulls \
+          -d "{\"title\":\"Update Aesop Gel to ${GEL_VERSION}\",\"head\":\"${BRANCH_NAME}\",\"base\":\"develop\"}"
```
maxlew, line 48: This is better than using the external action and should be fine for now, if this is something that we start relying on more or needing to change it might be worth moving this step to be done through a JS file or something that's a little easier for people to pick up and edit.

maxlew (commented):
This is starting to look really good now, a few little semantic comments from me. Trying to get it to line up as much as possible with a PR like this one 
https://github.com/aesop/aesop-graphql-service/pull/262

.github/workflows/update-webui.yml
MkMan: Done 👍

.github/workflows/update-webui.yml
MkMan: Still figuring out how to get current version, but the rest is done 👍

.github/workflows/update-webui.yml
MkMan: Done 👍

maxlew (approved):
Looks good! I guess most of the logic is now in the aesop-github-action repo so tag me on the review there when that's good for review :)

PR #172, opened 2020-09-10
[Feature][AES-1511] Create Product Extract Component for How to Use section [Fix for tests]

hassanraza0054:
Fix import statement in ProductExtract.spec.js and updated snapshots


src/components/ProductExtract/ProductExtract.spec.js
```diff
@@ -3,7 +3,7 @@ import { configure, shallow } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 import renderer from 'react-test-renderer';
 import ProductExtract from './ProductExtract';
-import ProductExtractFixture from './ProductExtract.fixture';
+import ProductExtractFixture from './ProductExtractfixture';
```
beau-charman, line 6: Could you instead rename the `ProductExtractfixture.js` file to ProductExtract.fixture.js` to keep them consistent?  Thank you!

beau-charman (changes requested):

src/components/ProductExtract/ProductExtract.spec.js
hassanraza0054: Sorry about that. Done!

beau-charman (approved):

PR #183, opened 2020-09-14
Run Github Actions on all branches

maxlew:
This will run the testing & linting actions on all branches, not just master & develop. It will also ignore any tag pushes (e.g releases)

The semantic release command also runs in this workflow however it will not make any changes unless on a 'release' branch (master or develop)


MkMan (approved):
Good idea

PR #186, opened 2020-09-14
Adding AddToCartCopy to HorizontalProductDisplayAccordion

maxlew:
Because of the way this component manages it's `product` state internally we can't update the `products` array after it's been sent to the component. To remedy this I've passed the `AddToCartCopy` as a separate prop that then gets passed into the add to cart button. 

Alternatively this component should have a useEffect that recalculates the `accordionProducts` state each time `props.products` changes.

### DIFF
--- a/src/components/HorizontalProductDisplayAccordion/HorizontalProductDisplayAccordion.js
+++ b/src/components/HorizontalProductDisplayAccordion/HorizontalProductDisplayAccordion.js
@@ -11,7 +11,7 @@ import { ascertainIsSmallOnlyViewport } from '~/utils/viewports';
 import useWindowHasResized from '~/customHooks/useWindowHasResized';
 import debounce from 'lodash/debounce';
 
-const HorizontalProductDisplayAccordion = ({ id, products }) => {
+const HorizontalProductDisplayAccordion = ({ id, products, addToCartCopy }) => {
   const [accordionProducts, toggleAccordionProducts] = useState(products);
   const [accordionActive, toggleAccordionActiveState] = useState(false);
   let isMobile = ascertainIsSmallOnlyViewport();
@@ -81,7 +81,7 @@ const HorizontalProductDisplayAccordion = ({ id, products }) => {
     return function callback() {
       window.removeEventListener('resize', resetAccordionOnResize);
     };
-  });
+  }, []);
 
   useWindowHasResized();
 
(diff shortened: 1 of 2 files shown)


src/components/HorizontalProductDisplayAccordion/components/AccordionProduct/AccordionProduct.js
```diff
@@ -20,6 +20,7 @@ const AccordionProduct = ({
   resetAccordion,
   toggleAccordion,
 }) => {
+  console.log('AccordionProduct', addToCartCopy);
```
hassanraza0054, line 23: Do we want to keep the console.log for debugging purposes? Just confirming.

src/components/HorizontalProductDisplayAccordion/HorizontalProductDisplayAccordion.js
```diff
@@ -81,7 +81,7 @@ const HorizontalProductDisplayAccordion = ({ id, products }) => {
     return function callback() {
       window.removeEventListener('resize', resetAccordionOnResize);
     };
-  });
+  }, []);
```
MkMan, line 84: I thought this would give linting errors because it doesn't include variables used in the body of the function 🤔

maxlew: There is a linting warning already on `isMobile = ascertainIsSmallOnlyViewport();` already. We only return a warning for `react-hooks/exhaustive-deps` as well so that's why this is still passing.

Not sure I really want to start fixing this as it might be only working as it's broken, it should just be adding and removing the resize listener but the function that's called is super side-effecty and changing state + local variables outside of a render process. 
```
        accordionProducts.map(product => {
          product.isExpanded = false;
          product.isCompressed = false;
        });
        isMobile = ascertainIsSmallOnlyViewport()
```

src/components/HorizontalProductDisplayAccordion/components/AccordionProduct/AccordionProduct.js
maxlew: good catch, removed this now

hassanraza0054 (approved):

src/components/HorizontalProductDisplayAccordion/HorizontalProductDisplayAccordion.js
MkMan: Fair enough

PR #193, opened 2020-09-17
[Feature][AES-1511] Create Product Extract Component for How to Use section [Add Flex for image]

hassanraza0054:
Included Flex logic to make image with dynamic while setting max height to 50px as per requirement.

### DIFF
--- a/src/components/ProductExtract/ProductExtract.js
+++ b/src/components/ProductExtract/ProductExtract.js
@@ -23,7 +23,6 @@ const ProductExtract = forwardRef(function ProductExtractRef(
     styles.base,
     styles[theme],
     {
-      [styles.hasBottomBorder]: hasBottomBorder,
       [styles.slideIn]: isVisible,
     },
     styles[`item${itemNum}`],
@@ -31,6 +30,11 @@ const ProductExtract = forwardRef(function ProductExtractRef(
   );
 
   const linkClassSet = cx(styles.link);
+
+  const flexClassSet = cx(styles.test, styles.delayAnimation, {
+    [styles.hasBottomBorder]: hasBottomBorder,
+    [styles.slideIn]: isVisible,
+  });
 
   const headingClassSet = cx(
     styles.heading,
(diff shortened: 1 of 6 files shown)


zack76 (changes requested):
Screen Shot 2020-09-17 at 4 53 54 pm
One border is missing, could you double-check with it?

Also the width of this slice looks more than 100%

hassanraza0054:
> Screen Shot 2020-09-17 at 4 53 54 pm
> 
> One border is missing, could you double-check with it?
> Also the width of this slice looks more than 100%

Ah the DefinitionListWithProductExtract file is just a showcase component. If you want to see the borders, just tick the relevant boxes in the knobs! 

Let me have a look at the width issue.

hassanraza0054:
Just adding a note that the width seems to be alright from my end. Made another commit to correct fixtures and add image properties as per your suggestion. Thanks @zack76

src/components/ProductExtract/__snapshots__/ProductExtract.spec.js.snap
```diff
@@ -11,36 +11,50 @@ exports[`<ProductExtract /> renders base component correctly 1`] = `
...
-    <h4
-      className="base xXSmall dark heading delayAnimation slideIn"
-    >
-      <span>
-        Works well with
-      </span>
-    </h4>
     <div
-      className="productName delayAnimation slideIn"
+      className="test delayAnimation slideIn"
```
kevin-ho87, line 15: Where is this `test` coming from?

src/components/ProductExtract/ProductExtract.js
```diff
@@ -32,6 +31,11 @@ const ProductExtract = forwardRef(function ProductExtractRef(
 
   const linkClassSet = cx(styles.link);
 
+  const flexClassSet = cx(styles.test, styles.delayAnimation, {
```
zack76, line 34: `test` here? @kevin-ho87 
should be deleted i think @hassanraza0054

zack76 (changes requested):

src/components/ProductExtract/__snapshots__/ProductExtract.spec.js.snap
beau-charman:
```
.test {
  display: flex;
  justify-content: space-between;
}
```

hassanraza0054: Change made! Sorry for that and thanks for the catch!

src/components/ProductExtract/ProductExtract.js
hassanraza0054: Done! Thanks for the catch

zack76 (approved):

kevin-ho87 (approved):

PR #197, opened 2020-09-18
[Feature][Feature/AES-1789] IconLink

zack76:
**New components: Icon Link for further use in Footer component**

Screen Recording 2020-09-18 at 1 49 08 pm


src/components/IconLink/IconLink.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+IconLink.defaultProps = {
+  altText: undefined,
+  className: undefined,
+  height: 21,
+  icon: undefined,
+  iconImageClassName: undefined,
+  id: undefined,
+  openInANewWindow: true,
```
beau-charman, line 56: can we change the prop to `hasTargetInNewWindow` for boolean naming and Hyperlink prop name consistency?

src/components/IconLink/IconLink.js
```diff
@@ -0,0 +1,61 @@
...
+      title={altText}
+      url={target}
+    >
+      <Image
+        altText={altText}
+        className={iconImageClassName}
+        large={icon.small}
+        medium={icon.small}
+        small={icon.small}
+        style={{ height: `${height}px`, width: `${width}px` }}
```
beau-charman, line 31: Do these values need to change with media queries?

beau-charman (changes requested):

src/components/IconLink/IconLink.js
zack76: updated

src/components/IconLink/IconLink.js
zack76: updated

kevin-ho87:
Just a curious question. what happens when different images with different heights are added? how would they line up?

kevin-ho87 (approved):

PR #201, opened 2020-09-21
fix: enable video autoplay

MkMan:
Enabling autoplay as videos were not showing up without it.

### DIFF
--- a/src/components/DoubleMedia/components/MediaBlock/MediaBlock.js
+++ b/src/components/DoubleMedia/components/MediaBlock/MediaBlock.js
@@ -27,7 +27,7 @@ const MediaBlock = ({
       <Video
         fallbackImage={fallbackImage}
         hasAllowAudio={false}
-        hasAutoplay={isScrollBasedVideo ? false : true}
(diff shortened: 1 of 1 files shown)


src/components/DoubleMedia/components/MediaBlock/MediaBlock.js
```diff
@@ -27,7 +27,7 @@ const MediaBlock = ({
       <Video
         fallbackImage={fallbackImage}
         hasAllowAudio={false}
-        hasAutoplay={isScrollBasedVideo ? false : true}
+        hasAutoplay={true}
```
andrwchang, line 30: was this always set to on initially?

MkMan: Yeah, I made this change in my previous PR.

andrwchang: yeap.. I checked

andrwchang (approved):

PR #208, opened 2020-09-24
Feature/aes 1557 create navigation bar

MkMan:
This is to add the navigation bar similar to the one [here ](https://www.aesop.com/au/c/body-hand/kits/) to Gel.

Navigation bar

### DIFF
--- a/src/components/NavigationBar/NavigationBar.js
+++ b/src/components/NavigationBar/NavigationBar.js
@@ -1,20 +1,94 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
+import Hyperlink from '~/components/Hyperlink';
 import styles from './NavigationBar.module.css';
 
-const NavigationBar = ({ className }) => {
-  const classSet = cx(styles.base, className);
+const testRefs = {
+  wrapper: 'NAVIGATION_BAR',
+  parentLink: 'NAVIGATION_BAR_PARENT_LINK',
+  childLink: 'NAVIGATION_BAR_CHILD_LINK',
+};
 
-  return <div className={classSet}>Test</div>;
+const NavigationBar = ({
+  childLinks,
+  className,
+  parentLink,
+  selectedUrl,
+  theme,
+}) => {
+  const hasChildren = !!childLinks.length;
+
+  if (!hasChildren && !parentLink) {
+    return null;
+  }
+
+  const links = [...childLinks];
+
+  if (parentLink) {
+    links.splice(0, 0, parentLink);
+  }
+
+  return (
+    <div
+      className={cx(styles.wrapper, className)}
+      data-test-ref={testRefs.wrapper}
+    >
+      <ul className={cx(styles.list)}>
+        {links.map(({ hasTargetInNewWindow, text, url }, index) => {
+          const linkClasses = cx({
+            [styles.link]: true,
+            [styles.hasChildren]: hasChildren && parentLink && index === 0,
+            [styles.isActive]: url === selectedUrl,
+          });
+
+          return (
+            <li key={index}>
+              <Hyperlink
+                className={linkClasses}
+                dataTestRef={
+                  parentLink && index === 0
+                    ? testRefs.parentLink
+                    : testRefs.childLink
+                }
+                hasTargetInNewWindow={hasTargetInNewWindow}
+                theme={theme}
+                url={url}
+              >
+                <span className={cx(styles.linkInnerText)}>{text}</span>
+              </Hyperlink>
+            </li>
+          );
+        })}
+      </ul>
+    </div>
+  );
 };
 
 NavigationBar.propTypes = {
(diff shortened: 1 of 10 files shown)


src/components/NavigationBar/NavigationBar.spec.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import { configure, shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import NavigationBar from './NavigationBar';
+
+configure({ adapter: new Adapter() });
+
+describe('<NavigationBar />', () => {
```
zack76, line 8: love the way how you write test cases

MkMan: Thank you 😃 🙏

src/components/NavigationBar/NavigationBar.js
```diff
@@ -0,0 +1,105 @@
...
+
+          return (
+            <li key={index}>
+              <button
+                className={linkClasses}
+                data-test-ref={testRefs.childLink}
+                onClick={() => selectFn(target)}
+              >
+                <span className={cx(styles.linkInnerText)}>{name}</span>
+              </button>
```
kevin-ho87, line 70: I think it might be better to use an `<a />` tag for semantic/accessibility purposes as it provides a page target. Even though the current live uses buttons. Snippet below is what is used for the new Gift Guide pages.
https://github.com/aesop/aesop-web-ui/blob/feature/aes-1590--gift-guide/src/app/components-custom/category-content-section/FilterSubNav.js
```
              <a
                className={categoryClassnames(cat.id)}
                href={`${mainCategory.url}${cat.id}/`}
                onClick={e => handleLinkClick(e, cat)}
                data-test-ref={testRefs.NAV_CATEGORY}
              >
                <span className="FilterBarSubnav__linkInnerText">
                  {cat.name || cat.id}
                </span>
              </a>
```

src/components/NavigationBar/NavigationBar.module.css
```diff
@@ -0,0 +1,107 @@
...
+
+  &::before {
+    left: 0;
+    width: 20px;
+    background-image: linear-gradient(to right, #fff 0%, transparent 100%);
+  }
+  &::after {
+    right: 0;
+    width: 50px;
+    background-image: linear-gradient(to right, transparent 0%, #fff 100%);
```
kevin-ho87, line 23: Are you able to use the same value instead of `transparent`. eg: 
```
    background: linear-gradient(
      to right,
      rgba(235, 234, 222, 1) 0%,
      rgba(235, 234, 222, 0) 100%
    ); /* W3C, IE10+, FF16+, Chrome26+, Opera12+, Safari7+ */
```
IOS has an issue of creating a white background.

zack76 (commented):
looks a bit weird when the background color is set to dark. Is this because the style requirement?
Screen Shot 2020-09-25 at 2 26 26 pm

zack76:
> looks a bit weird when the background color is set to dark. Is this because the style requirement?
> Screen Shot 2020-09-25 at 2 26 26 pm

oh, i just notice that it also happens on the current prod site

src/components/NavigationBar/NavigationBar.js
MkMan: Good point 👍

MkMan:
> looks a bit weird when the background color is set to dark. Is this because the style requirement?
> Screen Shot 2020-09-25 at 2 26 26 pm

Yeah, I came across that while integrating it with Web UI. I am looking into a way to use the background color value instead of the hardcoded white.

src/components/NavigationBar/NavigationBar.js
MkMan: I have changed it 🙂 Please have a look and if it's all right mark this resolved 🙏

src/components/NavigationBar/NavigationBar.js
```diff
@@ -0,0 +1,110 @@
...
+
+  return (
+    <div
+      className={cx(styles.wrapper, className, styles[theme])}
+      data-test-ref={testRefs.wrapper}
+    >
+      <ul className={cx(styles.list)}>
+        {parentLink && parentLink.name && (
+          <li>
+            <NavigationLink
```
kevin-ho87, line 38: Would it help to use the existing `hyperlink` component? https://aesop.github.io/aesop-gel/?path=/story/components-hyperlink--base-component. Then maybe add an onClick handler if there needs to be a js override

MkMan: Done :+1:

src/components/NavigationBar/NavigationBar.module.css
MkMan: Done :+1:

zack76 (approved):

beau-charman (approved):

PR #237, opened 2020-10-16
feat: ck-20 add to podium and loading stories

beau-charman:
This PR makes adjustments to the Podium and Loading components to address https://aesoponline.atlassian.net/browse/CK-20


src/components/Transition/Transition.module.css
```diff
@@ -2,268 +2,300 @@
...
 
 .fadeExitDone {
   opacity: 0;
+  transition: opacity 500ms linear;
 }
 
 .shiftInDownEnter {
   opacity: 0;
   transform: translate3d(0, -20px, 0);
+  transition:
```
MkMan, line 34: Is it worth making a mixin out of this? I noticed it gets added multiple times here.

src/components/Loading/Loading.module.css
```diff
@@ -32,6 +37,16 @@
     height: 5px;
   }
 
+  .medium & {
+    width: 7px;
```
MkMan, line 41: If the dimensions in `.dot` are meant to be the same as the ones in `.small` `.medium` `.large` should they be referencing a variable so the element doesn't break if someone changes one without updating the other?

MkMan (approved):

PR #241, opened 2020-10-20
feat(hooks): add  useExecuteOnImpression hook

MkMan:
Adding a hook to execute a function when the element comes into view.

hook


zack76 (approved):

hassanraza0054 (approved):

src/customHooks/useExecuteOnImpression/useExecuteOnImpression.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A hook to execute a callback when the element enters the fold
+ * @param {HTMLElement} ref a DOM reference to the element
+ * @param {Function} functionToExecute a callback function to execute
+ * @param {{threshold: Number, shouldExecuteOnReEntry: Boolean}} options hook behaviour options
+ *
+ *
+ * @example
+ * const ref = React.useRef(null);
+ * const myFunction = () => console.log('Execute order 66');
```
beau-charman, line 13: 😂

beau-charman: tenor

beau-charman (changes requested):
The only thing I would ask to change is the references to `functionToCall`, `functionToExecute`, etc to `callback` - to communicate what it actually is.

beau-charman:
Like the verb `execute` is cool and all, but the use of it makes it seem that it's doing something special within this hook - like a event handler has access to its target. But this just calls a callback function. So calling the argument `callback` will communicate that. I hope that makes sense

beau-charman (approved):

PR #248, opened 2020-10-28
Feature/aes 2079 internal promo tracking

MkMan:
This PR wraps the following components in `forwardRef`:
* Accordion
* BodyCopy
* Subnav
* TwoColumnList

Adds an `onClick` prop to the HyperLink component and passes it down in other places where it's needed for this feature to work.

### DIFF
--- a/src/components/Accordion/Accordion.js
+++ b/src/components/Accordion/Accordion.js
@@ -13,55 +13,59 @@ import Icon from '~/components/Icon';
 import { isObjectPopulatedArray } from '~/utils/objects';
 import styles from './Accordion.module.css';
 
-const Accordion = forwardRef(({ className, items, theme }, ref) => {
-  const [activeNodes, setIsActiveNodes] = React.useState([]);
+const Accordion = forwardRef(
+  ({ className, items, theme, wrapperClass }, ref) => {
+    const [activeNodes, setIsActiveNodes] = React.useState([]);
 
-  if (!isObjectPopulatedArray(items)) {
-    return null;
-  }
+    if (!isObjectPopulatedArray(items)) {
+      return null;
+    }
 
-  const classSet = cx(styles.base, styles[theme], className);
-  const handleOnChange = nodes => setIsActiveNodes(nodes);
-  const checkIsActive = id => activeNodes.includes(id);
+    const classSet = cx(styles.base, styles[theme], className);
+    const handleOnChange = nodes => setIsActiveNodes(nodes);
+    const checkIsActive = id => activeNodes.includes(id);
 
-  return (
-    <div ref={ref}>
-      <AccessibleAccordion
-        allowMultipleExpanded={true}
-        allowZeroExpanded={true}
-        className={classSet}
-        onChange={handleOnChange}
-      >
-        {items.map(({ content, heading, id }) => (
-          <AccordionItem className={styles.item} key={id} uuid={id}>
-            <AccordionItemHeading className={cx(styles.heading, styles[theme])}>
-              <AccordionItemButton className={styles.button}>
-                <span className={styles.headingCopy}>{heading}</span>
-                <Icon
-                  className={styles.icon}
-                  height={16}
-                  name="downArrow"
-                  theme={theme}
-                  width={16}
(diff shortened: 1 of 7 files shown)


zack76 (approved):

src/components/Accordion/Accordion.js
```diff
@@ -25,39 +25,43 @@ const Accordion = ({ className, items, theme }) => {
...
-              isOpened={checkIsActive(id)}
-              theme={{ collapse: styles.collapse }}
-            >
-              <div className={cx(styles.content, styles[theme])}>{content}</div>
-            </Collapse>
-          </AccordionItemPanel>
-        </AccordionItem>
-      ))}
-    </AccessibleAccordion>
+    <div ref={ref}>
```
kevin-ho87, line 28: I wonder if wrapping this component with a div will affect where its being used in web-ui

MkMan: In what way 🤔 ? Do you know what I should look for to check nothing broke?

src/components/Hyperlink/Hyperlink.js
```diff
@@ -17,6 +17,7 @@ const Hyperlink = ({
   hasTargetInNewWindow,
   isAlternate,
   isDownload,
+  onClick,
```
zack76, line 20: i was thinking if it should be named as `handleClick` or `handleOnClick`

src/components/Accordion/Accordion.js
kevin-ho87: Maybe just have a quick look at any existing pages on web-ui where its being used. from my experience, a simple wrapping div can break a layout unintentionally.

kevin-ho87: eg: if the holder in web-ui has a `display: grid`, the children could collapse.

MkMan: Good point, I'll have a look.

src/components/Hyperlink/Hyperlink.js
MkMan: How come? 🙂 
The prop is the `onClick` function and gets assigned to the `onClick` attribute without any manipulation in between. onClick is the expected name for this prop and it might be confusing to give it a different name.

zack76: i remember Beau mentioned before when defining an event handler, prefix with handle, so handleOnClick. But it's just passing a func as a prop here, i think you don't need to worry about it. Sorry for misunderstanding

src/components/Accordion/Accordion.js
MkMan: Checked on https://www.aesop.com/au/r/online-offering/ and it looks the same on desktop and mobile 🙌 

### On prod
Screen Shot 2020-10-29 at 10 18 04 am


### On this branch
After changes

kevin-ho87: cool! 👍 . Maybe just add a `className` prop that can be passed through to this holding div so that i can be styled if needed.

MkMan: I added a `wrapperClass` prop, thanks for the suggestion 👍

src/components/Hyperlink/Hyperlink.js
beau-charman: props don't have the `handle` prefix, props follow the HTML/JS convention. This keeps an easier to read API as props / contexts and functions are passed around. The function declared that fires when the `onClick` event is satisfied does have the `handle` prefix to clearly denote it is an event callback: 

```

const useMouseEvents = () => {
  const handleOnMouseOver = () => { ... }
  ...
  return {
    onMouseOver: handleOnMouseOver,
  }
}

...

const Component = ({ onFocus }) => {
  const { onMouseOver } = useMouseEvents();
  const handleOnClick = () => { ... }

  return <OtherComponent
    onClick={handleOnClick}
    onFocus={onFocus}
    onMouseOver={onMouseOver}
```

kevin-ho87 (approved):

beau-charman (approved):

PR #249, opened 2020-10-28
[Feature][AES-2163] Add Background Image Option to Full Screen Scroll Component

beau-charman:
This PR addresses https://aesoponline.atlassian.net/browse/AES-2163 by adding a background element component, styles and logic to display a video or image given props provided

### DIFF
--- a/.storybook/GridOverlay.js
+++ b/.storybook/GridOverlay.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import cx from 'classnames';
 import styles from './GridOverlay.module.css';
 
@@ -12,11 +13,13 @@ const generateColumns = () => {
   return columns;
 };
 
-const GridOverlay = ({ isActive = false }) => {
+const GridOverlay = ({ hasInvertedColours = false, isActive = false }) => {
   if (!isActive) return false;
 
   return (
-    <div className={styles.base}>
+    <div
+      className={cx(styles.base, { [styles.invertColour]: hasInvertedColours })}
+    >
       <div className={cx(styles.gutter, styles.left)} />
       {generateColumns()}
       <div className={cx(styles.gutter, styles.right)} />
@@ -24,8 +27,9 @@ const GridOverlay = ({ isActive = false }) => {
   );
 };
 
-
-
-GridOverlay.propTypes = {}
+GridOverlay.propTypes = {
+  hasInvertedColours: PropTypes.bool,
+  isActive: PropTypes.bool,
+};
 
 export default GridOverlay;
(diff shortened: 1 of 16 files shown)


.storybook/GridOverlay.js
```diff
@@ -12,20 +13,23 @@ const generateColumns = () => {
   return columns;
 };
 
-const GridOverlay = ({ isActive = false }) => {
+const GridOverlay = ({ hasInvertedColours = false, isActive = false }) => {
```
MkMan, line 16: I have seen other components setting the default prop values using `defaultProps` after the prop type definitions. Should it be the same here?

src/components/FullWidthHeroScroll/components/BackgroundElement/BackgroundElement.js
```diff
@@ -0,0 +1,41 @@
...
+      />
+    );
+  }
+
+  return null;
+};
+
+BackgroundElement.propTypes = {
+  image: PropTypes.object,
+  imageOrVideoAsBackground: PropTypes.oneOf(['image', 'video']),
```
MkMan, line 31: Would be better to name this `mediaType` so it's future proof in case a different type is to be used as background? I can't really think of other types at this stage but thought I'd bring up the idea.

beau-charman: Agree, I am trying hard to juggle the Contentful API with the GEL API, but I guess I have already used 'Media' else where. I will change.

.storybook/GridOverlay.js
beau-charman: Thats is true, but that is only to get the automated documents to show in storybook. I am more of a fan of the default arguments rather than default props, and this is a storybook helper, not a GEL component.

MkMan: Makes sense 👍

src/components/FullWidthHeroScroll/FullWidthHeroScroll.js
```diff
@@ -2,15 +2,18 @@ import React, { useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { HYPERLINK_STYLE_TYPES } from '~/constants';
-import { BodyCopy, Hyperlink, Icon, Video } from '~/components';
+import { BodyCopy, Hyperlink, Icon } from '~/components';
 import FullscreenSection from './components/FullscreenSection';
+import BackgroundElement from './components/BackgroundElement';
```
zack76, line 7: Maybe the order here is not perfect although it's working

MkMan (approved):

PR #261, opened 2020-11-13
fix(deps): remove circular dependencies

MkMan:
I am a bit conflicted on this. We get warnings of circular dependencies but things are working fine.
image

I made this change to eliminate those circular dependencies which makes some imports a bit ugly but 🤷‍♂️
image


beau-charman:
any chance we could just keep the component imports as `~/components/..` rather than using the relative imports?

beau-charman (approved):

MkMan:
> any chance we could just keep the component imports as `~/components/..` rather than using the relative imports?

Good point 👌  I updated the imports

PR #262, opened 2020-11-16
AES-1481: enable native lazy loading on image component

maxlew:
Really simple change, but updates quite a few spec files. This PR enables native lazy loading for https://aesoponline.atlassian.net/browse/AES-1481

Lazy loading is enabled by default on the image component. It can be disabled by passing `isLazyLoaded={false}`

Can see some more info about native implementation here https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading


src/components/Image/Image.js
```diff
@@ -7,7 +7,18 @@ import Hyperlink from '~/components/Hyperlink';
 import styles from './Image.module.css';
 
 const Image = forwardRef(function ImageRef(
-  { altText, className, cta, large, medium, small, style, type, theme },
+  {
+    altText,
+    className,
+    cta,
+    isLazyLoaded,
```
MkMan, line 14: Great idea allowing the consumer the ability to set it 👌

MkMan (approved):

MkMan:
When you see that this feature isn't supported in IE
Mild shock

It is interesting though that Safari doesn't support it, it's also still in the experimental stage. We may need to revisit it in the future but I think for now it's good to have.

beau-charman:
Interested to see how this behaves in a feature branch, especially with all the CSS Transitions Group usage, if `loading="lazy"` will work with image tags that are not mounted yet.

MkMan:
> Interested to see how this behaves in a feature branch, especially with all the CSS Transitions Group usage, if `loading="lazy"` will work with image tags that are not mounted yet.

@beau-charman I found this link on the Jira ticket, It still works 
https://aes-1481--lazy-loading.acceptance.aesop.com/au/r/skin-care/how-to-clease/?contentful=true

maxlew:
> It is interesting though that Safari doesn't support it, it's also still in the experimental stage. We may need to revisit it in the future but I think for now it's good to have.

The implementation is pretty different between Firefox and Chrome. Chrome will still load images when they are a few screen heights away and seems to do some tricks around network speed to increase or decrease that. Firefox just loads them when they come into view which doesn't look the best. 

I kinda hate it but Google are pushing this heavily as it's now part of their core web vitals, with the added thread that it'll affect search rankings in the near future.

beau-charman:
Hey @maxlew , do you know if feature branches run slower than the base acceptance endpoint?

beau-charman:
I did hit reload on aes-1481--lazy-loading.acceptance like half a second later, but it loads far slower.

beau-charman:
The lazy loading works well though, as you scroll the page the assets load in when needed in the network tab, just hoping the long load is a feature branch side effect

beau-charman:
I am going to say feature branches are weaker. 

Stat-wise `aes-1481--lazy-loading.acceptance` is sending far less on initial load: 

Screen Shot 2020-11-18 at 9 52 01 am

compared to acceptance: 

Screen Shot 2020-11-18 at 9 52 36 am

Happy to see this one tested in acceptance proper.

beau-charman (approved):
Would like to see tests ran on https://acceptance.aesop.com and compared to current state

maxlew:
> Hey @maxlew , do you know if feature branches run slower than the base acceptance endpoint?

Yeah the feature branches don't run through our Akamai CDN. So including some other things there's no caching or auto image optimization happening

PR #263, opened 2020-11-18
[Feature][AES-2182] PDP - GraphQL service and web-ui GraphQL Schema updates

beau-charman:
This PR makes changes to the PDP Header and Add To Cart components to address https://aesoponline.atlassian.net/browse/AES-2182

- Add is `isSellable` to Add To Cart Components
- Move `cartDisclaimer` to variants level
- Add `size` variants - `singular` and `plural`
- Conditionally add ingredients row if applicable 
- Update image transition hook

### DIFF
--- a/src/components/ProductDetailHeader/ProductDetailHeader.js
+++ b/src/components/ProductDetailHeader/ProductDetailHeader.js
@@ -56,7 +56,18 @@ const ProductDetailHeader = ({ breadcrumbs, className, copy, theme }) => {
 };
 
 ProductDetailHeader.propTypes = {
-  breadcrumbs: PropTypes.object,
+  breadcrumbs: PropTypes.shape({
+    className: PropTypes.string,
+    items: PropTypes.arrayOf(
+      PropTypes.shape({
+        label: PropTypes.string,
+        id: PropTypes.string,
+        url: PropTypes.string,
+        title: PropTypes.string,
+      }),
+    ),
+    theme: PropTypes.oneOf(['dark', 'light']),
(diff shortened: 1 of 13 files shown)


src/components/ProductDetailHeader/ProductDetailHeader.js
```diff
@@ -79,11 +93,18 @@ ProductDetailHeader.propTypes = {
 };
 
 ProductDetailHeader.defaultProps = {
-  breadcrumbs: undefined,
+  breadcrumbs: {
+    className: undefined,
+    items: undefined,
```
MkMan, line 98: Would having this as `[]` eliminate the need for falsy checks before calling array functions?

src/components/ProductExtract/ProductExtract.js
```diff
@@ -113,8 +113,8 @@ const ProductExtract = forwardRef(function ProductExtractRef(
           <div className={productImageClassSet}>
             <Image
               altText={product.image.altText}
-              large={product.image.small}
-              medium={product.image.small}
+              large={product.image.large}
```
MkMan, line 116: Good pick up 👌

MkMan (approved):

src/components/ProductDetailHeader/ProductDetailHeader.js
beau-charman: I've leave for now as the logic in Breadcrumbs is already there - the proptype warning for Array should discourage any other value being passed also

PR #271, opened 2020-11-23
chore: remove deprecated set-env call

MkMan:
Set env variable using the new specified way [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable)

I tried this with a test tag and you can see the logs [here](https://github.com/aesop/aesop-gel/runs/1439647915?check_suite_focus=true). I cancelled the job before it raised a PR on web ui.

image

### DIFF
--- a/.github/workflows/update-webui.yml
+++ b/.github/workflows/update-webui.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: Create variables
         run: |
-          echo "::set-env name=GEL_VERSION::${GITHUB_REF#refs/*/}"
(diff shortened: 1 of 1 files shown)


maxlew (approved):
Looks good, I think technically we're still leaking the ENV value. But it's a git ref so no issue

MkMan:
> Looks good, I think technically we're still leaking the ENV value. But it's a git ref so no issue

I'll merge this for now to get things working again but what do you think is a good way to address that?

PR #272, opened 2020-11-23
fix: incorrect window check

MkMan:
The `isBrowser` check was only included for the first condition, this change makes it before all conditions.


beau-charman (approved):
also makes sense to short-circuit on `isInBrowser() ...`

PR #278, opened 2020-11-27
chore: remove unused dependencies

MkMan:
I noticed these 2 dev dependencies but found no references to them in the repo.


package.json
```diff
@@ -121,7 +121,6 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^26.0.1",
     "jest-cli": "^26.0.1",
-    "npm-merge-driver": "^2.3.6",
```
hassanraza0054, line 124: Not sure but this seems to be a dev tool rather than something explicitly used in the repo.

https://www.npmjs.com/package/npm-merge-driver

If my understanding isn't wrong, maybe it'd be good to get others feedback as to whether they use this tool?

@maxlew actually curious to hear your thoughts on this since we've talked about package.lock conflicts before.

package.json
```diff
@@ -149,7 +148,6 @@
     "style-loader": "^1.1.1",
     "stylelint": "^12.0.1",
     "stylelint-config-css-modules": "^2.1.0",
-    "stylelint-config-idiomatic-order": "^8.1.0",
-    "webpack-cli": "^3.3.12"
```
hassanraza0054, line 153: Same idea here. Seems to be a tool rather than something meant to be directly referenced in the repo no?

https://www.npmjs.com/package/webpack-cli

package.json
MkMan: If it's something not used by the repo it shouldn't be included as part of its dependencies. Instead, the dev using it can install it globally on their machine.

package.json
MkMan: webpack cli is used to operate or work with webpack. Gel doesn't use webpack though, it uses `rollup`.

hassanraza0054: Thanks for that! Learning a lot. Appreciate the work Mustafa. All good from me

package.json
hassanraza0054: Didn't see it that way but in hindsight you're definitely right! Thanks for the clarity Mustafa

hassanraza0054 (approved):

PR #279, opened 2020-11-30
fix: update carousel prop types

MkMan:
Most of the time the slide description is just a `string`, however, sometimes it's a react node. This change allows the component to accept both

### DIFF
--- a/src/components/Carousel/Carousel.js
+++ b/src/components/Carousel/Carousel.js
@@ -233,7 +233,7 @@ Carousel.propTypes = {
       url: PropTypes.string,
       text: PropTypes.string,
     }),
-    description: PropTypes.string,
+    description: PropTypes.node,
     eyebrow: PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/components/Carousel/Carousel.js
```diff
@@ -233,7 +233,7 @@ Carousel.propTypes = {
       url: PropTypes.string,
       text: PropTypes.string,
     }),
-    description: PropTypes.string,
+    description: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
beau-charman, line 236: save some characters, make all of theses `description: PropTypes.node,`

beau-charman (changes requested):

src/components/Carousel/Carousel.js
MkMan: Done 👍

beau-charman (approved):

Issue #49, opened 2020-05-22
Add React Hooks Testing Library


beau-charman:
Documentation: https://github.com/testing-library/react-hooks-testing-library

PR #280, opened 2020-11-30
[Feature][AES-2264] Unit Test Refactor

beau-charman:
This PR is the first in a series to address https://aesoponline.atlassian.net/browse/AES-2264

### Stage One

**Static testing refactor**
- Clean up of eslint and stylelint declarations and execution 
- Consistent eslintrc entries (use `"error"` etc…)
- Stylelint now fixes css issues rather than sharing responsibility with prettier
- Markdown and JSON files are now run through prettier

**API refactor**
- Favour named exports over default exports for a consistent and transparent API
- Declare exports at the bottom of files for quick look up of intended accessible methods and values
- Moved Hooks into their own folders
- Added an error for context consumers that are not wrapped in appropriate provider

**Test coverage improvement**
- `statements` 65 👉 82
- `branches` 50 👉 65
- `functions` 40 👉 68
- `lines` 65 👉 83

**Test runner helpers and assertion refactor**
- Remove Enzyme in favour for Testing Library (React|Hooks|DOM)
- Added Jest Watch Typeahead helper

**Initial test refactor**
- Made required changes to test suites
- Audio component refactored into children components and parent consumer component

**Addresses**
- https://github.com/aesop/aesop-gel/issues/49


src/components/Audio/Audio.fixture.js
```diff
@@ -1,4 +1,5 @@
-export default {
+/* istanbul ignore file */
```
MkMan, line 1: Would using one of these 2 eliminate the need to manually ignore files?
* [modulePathIgnorePatterns](https://jestjs.io/docs/en/configuration#modulepathignorepatterns-arraystring)
* [coveragePathIgnorePatterns](https://jestjs.io/docs/en/configuration#coveragepathignorepatterns-arraystring)

MkMan:
So much nicer having each hook in its own folder 👌

MkMan:
I was curious why did you go with 

```js
const a = 1;

const b = 2;

export { a, b };
```

instead of just

```js
export const a = 1;

export const b = 2;
```

setupTests.js
```diff
@@ -1,4 +1,5 @@
 import 'regenerator-runtime/runtime';
+import '@testing-library/jest-dom/extend-expect';
```
MkMan, line 2: I don't think I tried it before but looks like a good addition to save us writing our own utils for that stuff.

src/components/Audio/Audio.fixture.js
beau-charman: Thank you!

beau-charman:
> I was curious why did you go with
> 
> ```js
> const a = 1;
> 
> const b = 2;
> 
> export { a, b };
> ```
> 
> instead of just
> 
> ```js
> export const a = 1;
> 
> export const b = 2;
> ```

Hey @MkMan , while refactoring the API for testing, I also noticed a lot of exported functions that were not used (or even required by other files), therefore I am going to establish a pattern in GEL where exports are only declared at the bottom of a file, (just as imports, besides dynamic imports, are declared at the top of a file), this will allow for a cleaner API

src/components/Audio/Audio.fixture.js
beau-charman: actually, I did try these, could not get the reg ex / globs to work 😭

src/components/Audio/Audio.js
```diff
@@ -21,223 +17,62 @@ const Audio = ({
...
-          >
-            {artistName}
-          </Heading>
-          <Heading hasSerifFont={true} isFlush={true} level="4" size="medium">
-            {trackTitle}
-          </Heading>
-        </header>
-      )}
+    <article className={classSet} id={id}>
+      <Header artistName={artistName} trackTitle={trackTitle} />
```
hassanraza0054, line 47: I really like what you did with the Header here. Solid stuff

src/components/Audio/Audio.js
```diff
@@ -21,223 +17,62 @@ const Audio = ({
...
-              className={styles.downloadButtonIcon}
-              height={13}
-              name="download"
-              width={13}
-            />
-          </Hyperlink>
-        </div>
-      </footer>
-    </div>
+      <Footer
```
hassanraza0054, line 55: Same thing with the footer. Solid encapsulation

src/components/Audio/Audio.js
```diff
@@ -265,18 +100,18 @@ Audio.defaultProps = {
...
   className: undefined,
   copy: {
-    loading: 'Loading audio file',
-    play: 'Play',
-    pause: 'Pause',
-    downloadLabel: 'Download Music',
-    downloadTitle: 'Download Music',
-    seekForward: 'Seek Forward',
-    seekBackward: 'Seek Backward',
+    downloadLabel: undefined,
```
hassanraza0054, line 103: Curious how all the defaultProps are now undefined whereas each of them had values before. What sparked this change?

src/components/Audio/Audio.spec.js
```diff
@@ -1,32 +1,26 @@
...
+        copy={AudioFixture.copy}
+        hasAutoPlay={AudioFixture.hasAutoPlay}
+        id={AudioFixture.id}
+        progressColor={AudioFixture.progressColor}
+        trackTitle={AudioFixture.trackTitle}
+      />,
+    );
 
-    expect(tree).toMatchSnapshot();
+    expect(container).toMatchSnapshot();
```
hassanraza0054, line 24: Adding as many questions for my own understanding but hope its useful to the team too. So the difference between the original **renderer** function and **render** from _react testing library_ is that the latter outputs the JSON() directly, making in compatible for the snapshot function without needing to run **.toJSON()**?

src/components/Audio/components/Footer/Footer.js
```diff
@@ -0,0 +1,126 @@
...
+    seekForwardTitle: undefined,
+  },
+  duration: undefined,
+  isLoading: undefined,
+  isPlaying: undefined,
+  onSeekBackwardButtonClick: undefined,
+  onSeekForwardButtonClick: undefined,
+  onPlayPauseButtonClick: undefined,
+  progress: undefined,
+  progressColor: 'orange',
```
hassanraza0054, line 123: Is there a value in defining any of these default properties?

src/components/Audio/Audio.js
beau-charman: These values were there from development / awaiting final contentful API. No component should have a default prop of something in english, all copy should be served via translations, unless is a hidden label or system text. 

cc @hassanraza0054

src/components/Audio/Audio.js
beau-charman: ty :) I like components to be composable. have a single purpose that can be reused or reuse other ideas

src/components/AddToCartButton/AddToCartButton.stories.mdx
```diff
@@ -18,29 +18,36 @@ Requires `AddToCartContextProvider`, with a prop of `onClick`, which is a callba
...
+      );
     })()}
   </Story>
-</Preview>
+</Canvas>
 
 ## API
 
-<Props of={AddToCartButton} />
+<ArgsTable of={AddToCartButton} />
```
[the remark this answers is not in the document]
beau-charman: its a deprecation with the new versions of Storybook

src/components/AddToCartButton/AddToCartButton.stories.mdx
```diff
@@ -18,29 +18,36 @@ Requires `AddToCartContextProvider`, with a prop of `onClick`, which is a callba
...
+          >
+            <AddToCartButton theme={theme} copy={AddToCartButtonFixture.copy} />
           </VariantSelectContextProvider>
         </AddToCartContextProvider>
-      )
+      );
     })()}
   </Story>
-</Preview>
+</Canvas>
```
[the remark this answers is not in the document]
beau-charman: its a deprecation with the new versions of Storybook

src/components/Audio/components/PausePlayButton/PausePlayButton.js
```diff
@@ -0,0 +1,69 @@
...
+        theme="dark"
+        width={15}
+      />
+      <Icon
+        className={cx(
+          styles.iconPause,
+          { [styles.hidden]: !isPlaying },
+          { [styles.disabled]: isLoading },
+        )}
+        height={15}
```
hassanraza0054, line 36: I imagine this will always be 15 and thus the hardcoded value, but will there ever be a case for us to later customise our play/pause buttons for difference pages? Like maybe 'small', 'medium' or 'large' play buttons? Might be overthinking this, but if so, would you think there's a value in storing these variables in a separate common space, similar to how we store some of our colour CSS vars? Maybe its a lot more work for a case that might not come, but I'm just thinking from a generalisation standpoint whether its a potential area of code to centralise?

src/components/Audio/components/SeekBackwardButton/SeekBackwardButton.js
```diff
@@ -0,0 +1,47 @@
...
+      <Button
+        className={cx(styles.base, styles[progressColor])}
+        isEnabled={!isLoading}
+        isInline={true}
+        onClick={onClick}
+        title={copy.title}
+      >
+        <Icon
+          className={cx(styles.icon, { [styles.disabled]: isLoading })}
+          height={20}
```
hassanraza0054, line 20: Same question here as above with the play button. The play/pause button had a height/width of 15, whereas this one has a height and width of 20. Considering possible future customisations and the idea to keep everything at scale, perhaps its worth saving these height and width values of buttons in a separate file? Curious on the teams thoughts

src/components/Audio/Audio.js
```diff
@@ -1,14 +1,10 @@
...
-import moment from 'moment';
-import throttle from 'lodash/throttle';
-import { COLORS, HYPERLINK_STYLE_TYPES } from '~/constants';
-import Button from '~/components/Button';
-import Heading from '~/components/Heading';
-import Hyperlink from '~/components/Hyperlink';
-import Icon from '~/components/Icon';
-import Loading from '~/components/Loading';
+import { Loading } from '~/components/Loading';
+import { Header } from './components/Header';
```
hassanraza0054, line 5: Hope i'm not being too nitpicky here, but for consistency, would **~/components/Header** be better?

hassanraza0054: Same for footer below

src/components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -13,7 +13,7 @@ const Breadcrumbs = ({ className, items, theme }) => {
   const classSet = cx(styles.base, styles[theme], className);
 
   return (
-    <nav className={classSet}>
+    <nav className={classSet} data-testid="data-testid-Breadcrumbs">
```
hassanraza0054, line 16: Will **data-testid** be our new variable to hook onto when creating unit tests using the react-testing-library? Also, the naming convention we'll be going for will be "data-testid-{componentName}"?

src/components/Audio/Audio.js
hassanraza0054: That makes sense. Thanks for the context Beau!

src/components/Breadcrumbs/Breadcrumbs.js
beau-charman: `data-testid` should only be used if there is nothing else to hook onto (The idea behind the Testing Library suite is to test the code as a user interacts with it, rather than concerning tests with implementation details). 

I am still refactoring the actual tests, hoefully will drop testid for this component. 

In regards to the convention, I wan IDs to be unique, stand up as used for tesing and searchable (`data-testid-component`, or `data-testid-Component`). Any suggestions welcome. 

cc @hassanraza0054

src/components/Audio/Audio.js
beau-charman: they are relative imports, using child components specific to the implementation of Audio. 

cc @hassanraza0054

src/components/Audio/Audio.spec.js
beau-charman: the `container` variable desctructured from `render` is a dom representation of where the component (or provided HTML) was mounted, and doesnt need to be parsed to JSON to be used

cc @hassanraza0054

src/components/Audio/components/PausePlayButton/PausePlayButton.js
beau-charman: I am going to leave the comments regarding the specifics of the Audio component out of the review consideration, as they have already been committed as is and are available in web-ui. If you want to raise an issue that we can come back to in regards to the Audio components implementation and API, please go for it. 

@hassanraza0054 

This PR regards the refactoring of existing code to better suit the testing framework.

src/components/Breadcrumbs/Breadcrumbs.js
MkMan: I've personally found test refs/ids to be quite handy when unit testing. They're a nicer way to find an element than by using css selectors.

beau-charman: Totally agree, but definitely reach for roles and text first :)

.storybook/main.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 
 module.exports = {
   stories: [
-    '../src/**/*.stories.(js|mdx)',
+    '../src/**/*.stories.@(js|mdx)',
```
hassanraza0054, line 5: What does adding the '@' symbol do here? I have seen it in imports across gel but I believe it was before this change too, so i'm just curious

.storybook/storybook.module.css
```diff
@@ -48,6 +48,9 @@
 }
 
 :global {
+  .sidebar-header img {
+    height: 100%;
```
hassanraza0054, line 52: What was happening here that you had to explicitly change the height of the sidebar-header img to 100%?

jest.config.js
```diff
@@ -1,29 +1,35 @@
...
   coverageThreshold: {
     global: {
-      branches: 50,
-      functions: 40,
-      lines: 65,
-      statements: 65,
+      statements: 82,
+      branches: 65,
+      functions: 68,
+      lines: 83,
```
hassanraza0054, line 20: Could you contextualise what actions/behaviour these 4 coverage threshold lines enforce? Curious since you've raised the values. Wonder whether its because we have more components now and this covers that, or whether it represents a percentage of working code that should be present in each of the 4 levels.

src/components/Breadcrumbs/Breadcrumbs.js
hassanraza0054: @beau-charman could you elaborate as to what you mean by reaching to roles and text first? Just want to completely understand what you're suggesting we aim for.

hassanraza0054 (approved):

src/components/AddToCartButton/AddToCartButton.spec.js
```diff
@@ -1,70 +1,160 @@
...
+
+    expect(button).toBeInTheDocument();
+
+    let breadcrumbs = screen.queryByTestId('data-testid-loading');
+
+    expect(breadcrumbs).not.toBeInTheDocument();
+
+    userEvent.click(button);
+
+    breadcrumbs = await screen.findByTestId('data-testid-loading');
```
zack76, line 156: just curious what the difference between `queryByTestId` and `findByTestId`

MkMan: I've found [this page](https://testing-library.com/docs/react-testing-library/cheatsheet/#queries) to be pretty useful 🙂

src/components/Audio/components/SeekForwardButton/SeekForwardButton.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { Button } from '~/components/Button';
+import { Icon } from '~/components/Icon';
+import styles from './SeekForwardButton.module.css';
+
+const SeekForwardButton = React.memo(
```
zack76, line 8: I feel like this component looks same/similar to `SeekBackwardButton` if I didn't miss anything. Is that possible we could integrate these two?

src/components/AddToCartButton/AddToCartButton.spec.js
zack76: > I've found [this page](https://testing-library.com/docs/react-testing-library/cheatsheet/#queries) to be pretty useful 🙂
> 
> image
Thank you!

src/components/ContentHubArticle/ContentHubArticle.spec.js
```diff
@@ -1,38 +1,32 @@
...
-          isReadMore={false}
-          longTitle={ContentHubArticleFixture.longTitle}
-          onClick={ContentHubArticleFixture.onClick}
-          readingTime={ContentHubArticleFixture.readingTime}
-          uri={ContentHubArticleFixture.uri}
-          verticalThumbnail={ContentHubArticleFixture.verticalThumbnail}
-        />,
-      )
-      .toJSON();
+  it('should render base component correctly', () => {
```
zack76, line 11: I believe I should add more test cases to this component once I get time 😃

src/components/AddToCartButton/AddToCartButton.spec.js
hassanraza0054: Wonder what the significance of async vs non-async is here.

MkMan:
> **Static testing refactor**
> 
> * Clean up of eslint and stylelint declarations and execution
> * Consistent eslintrc entries (use `"error"` etc…)
> * Stylelint now fixes css issues rather than sharing responsibility with prettier
> * Markdown and JSON files are now run through prettier

Nice! I am hoping we can get to a stage where we only use `eslint`, `stylelint` and `prettier` without other smaller neglected packages that link them. Hopefully one day we can move those rules to [aesop-code-styles](https://github.com/aesop/aesop-code-styles) and use it across our projects for consistent code formatting and less setting up.
image


> **API refactor**
> 
> * Favour named exports over default exports for a consistent and transparent API
> * Declare exports at the bottom of files for quick look up of intended accessible methods and values
> * Moved Hooks into their own folders
> * Added an error for context consumers that are not wrapped in appropriate provider
> 
> **Test coverage improvement**
> 
> * `statements` 65 👉 82
> * `branches` 50 👉 65
> * `functions` 40 👉 68
> * `lines` 65 👉 83

Gooooood stuff 👌

src/components/AddToCartButton/AddToCartButton.spec.js
MkMan: I quickly skimmed through docs and found the following:
* only `findBy` and `findAllBy` are async
* they will try to find element(s) within 1s and return them or throw an error if the element isn't found. Probably useful if elements take time to become present in the DOM. More info [here](https://testing-library.com/docs/dom-testing-library/api-queries#findby)

.storybook/main.js
MkMan: The `stories` field defines which files Storybook should load. It takes an array of `globs` (`glob`: pattern matching string).
According to Storybook docs, it supports globing that conform to the `micromatch` syntax. Visiting `micromatch` [docs](https://github.com/micromatch/micromatch#extglobs) shows that `@` *Matches one of the given patterns*.

jest.config.js
MkMan: > Could you contextualise what actions/behaviour these 4 coverage threshold lines enforce?

`branches` are directly related to conditional logic. In test coverage context it means that conditional logic branches were executed in the process of running the tests. 
```js
// a conditional will introduce 2 branches. When the condition is true and when it isn't
if (x) {
  // branch 1 
} else {
  // branch 2
}

// ....

if (y) {
  // branch 1
}
// in the absence of an else block, the second branch is just that the if block doesn't execute
```

`functions` are just that 😃 In test coverage context it means that the function was executed in the process of running the tests.

`statements` and `lines` are similar and often you'll see they have a close threshold number. Check this [answer](https://github.com/gotwarlost/istanbul/issues/639#issuecomment-225632261) out.

> Wonder whether its because we have more components now and this covers that, or whether it represents a percentage of working code that should be present in each of the 4 levels.

It's the later 😃  The increase in numbers basically means "fail the test task if less than 82% of all statements are executed, or less than 65% of all branches are executed, etc..."

package.json
```diff
@@ -99,28 +98,31 @@
...
+    "eslint-plugin-jest-dom": "^3.3.0",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-react": "^7.17.0",
     "eslint-plugin-react-hooks": "^2.5.1",
+    "eslint-plugin-testing-library": "^3.10.1",
     "husky": "^4.2.5",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^26.0.1",
     "jest-cli": "^26.0.1",
+    "jest-watch-typeahead": "^0.6.1",
```
MkMan, line 125: Good one 👌 ! I used it before in a previous job and thought it was built into `jest` and was wondering why I didn't get suggestions here.

src/customHooks/useExecuteOnImpression/useExecuteOnImpression.js
```diff
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { useOnScreen } from '~/customHooks/useOnScreen';
+import { useOnScreen } from '~/customHooks';
```
MkMan, line 2: I ran the build locally and this change causes a circular dependency. Can you please check on your side as well and revert it if you get the same?

MkMan:
Does this close #49 ?

src/components/Audio/Audio.customHooks.js
```diff
@@ -0,0 +1,78 @@
+import { useEffect, useRef, useState } from 'react';
+import throttle from 'lodash/throttle';
+import WaveSurfer from 'wavesurfer.js';
```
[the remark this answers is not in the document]
beau-charman: I wasnt aware that we were looking at SSR until like a week ago, I can look at changing this though. 

Same with Hassan comments regarding The Audio component - this code has been here for a while and is already available in Web UI - So if you could create an issue to address this, that would be great.

package.json
beau-charman: agree, it should just be in the main package 🙌

beau-charman:
> Nice! I am hoping we can get to a stage where we only use `eslint`, `stylelint` and `prettier` without other smaller neglected packages that link them. Hopefully one day we can move those rules to [aesop-code-styles](https://github.com/aesop/aesop-code-styles) and use it across our projects for consistent code formatting and less setting up.

Totally agree, just refactoring for this project at the moment - happy to remove the specific dependancy and use aesop-code-styles after all this work is across the line

src/customHooks/useExecuteOnImpression/useExecuteOnImpression.js
beau-charman: Thank you, was distracted by all the 'A plugin is directly adding properties...' outputs :(

src/components/Audio/Audio.customHooks.js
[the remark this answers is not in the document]
MkMan: Fair enough 🙂  Currently, Gel works with SSR after I changed this import [here](https://github.com/aesop/aesop-gel/blob/develop/src/components/Audio/Audio.js#L43) so changing it back would regress. If you could please retain the change or check the it works with SSR that would be best.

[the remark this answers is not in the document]
beau-charman: oh! didnt have that change when I started this branch, I will make that change here too.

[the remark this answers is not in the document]
MkMan: Thank you 😊

src/components/Breadcrumbs/Breadcrumbs.js
beau-charman: > @beau-charman could you elaborate as to what you mean by reaching to roles and text first? Just want to completely understand what you're suggesting we aim for.

https://testing-library.com/docs/guide-which-query/
https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

.storybook/storybook.module.css
beau-charman: It was collapsing - a computed height of 0. Not sure if it was because of a style introduced in the new storybook or what

jest.config.js
beau-charman: https://medium.com/@krishankantsinghal/how-to-read-test-coverage-report-generated-using-jest-c2d1cb70da8b

src/components/AddToCartButton/AddToCartButton.spec.js
beau-charman: https://testing-library.com/docs/guide-which-query/
https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

Good guide would be:
- use `get*` when you expect to find the target.
- use `query*` when you don't expect to find the target.
- use `find*` when you expect to wait for the target to be available in the DOM (Asynchronously loaded in / fetched)

src/components/Audio/components/SeekForwardButton/SeekForwardButton.js
beau-charman: Addressed this

src/components/ContentHubArticle/ContentHubArticle.spec.js
beau-charman: all good!

zack76 (approved):

MkMan (approved):

PR #283, opened 2020-12-16
fix: Address Postcss warnings from Web-ui

MkMan:
This to address some warnings in web-ui.

### DIFF
--- a/src/components/Carousel/components/Slide/Slide.module.css
+++ b/src/components/Carousel/components/Slide/Slide.module.css
@@ -5,18 +5,6 @@
   font-family: var(--font-family-sans-serif-medium);
   font-size: rem(14px);
 
-  @media (--viewport-sm) {
-    font-size: rem(14px);
-  }
-
-  @media (--viewport-md) {
-    font-size: rem(14px);
-  }
-
-  @media (--viewport-lg) {
-    font-size: rem(14px);
-  }
-
   &.dark {
     color: var(--color-dark-copy);
   }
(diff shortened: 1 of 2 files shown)


src/components/Carousel/components/Slide/Slide.module.css
```diff
@@ -5,18 +5,6 @@
   font-family: var(--font-family-sans-serif-medium);
   font-size: rem(14px);
 
-  @media (--viewport-sm) {
```
MkMan, line 8: This variable isn't defined

image

src/components/NavigationBar/NavigationBar.module.css
```diff
@@ -1,12 +1,14 @@
 @import 'styles/index';
 
-.wrapper {
+:root {
```
MkMan, line 3: Complains about undefined variables. I dug through the codebase and found that variables are declared inside `:root` so I am doing the same here.

Screen Shot 2020-12-17 at 10 09 48 am

src/components/Carousel/components/Slide/Slide.module.css
kevin-ho87: I wonder if it was meant to be `--viewport-sm-only` 🤔

MkMan: It might be, I didn't really bother with it too much because it seemed redundant as line 6 already declares the same style.

kevin-ho87: true

kevin-ho87 (approved):

PR #285, opened 2020-12-21
[Chore] update storybook

beau-charman:


MkMan (approved):
Did the update tool change the imports automatically? That would be impressive!

beau-charman:
> Did the update tool change the imports automatically? That would be impressive!

nah, just a find and replace in `**/*.stories*`, but VSCode might be able to do that

zack76 (approved):

PR #287, opened 2020-12-24
chore: delete duplicate file

MkMan:
Running `npm start` fails because of this file. There is a similar looking file inside [AddToCartContext](https://github.com/aesop/aesop-gel/blob/081d14801fc3fdde084f47a312e4952742093bf0/src%2Fcontexts%2FAddToCartContext%2FAddToCartContext.stories.mdx#L0-L1), so I am guessing this one was missed during the refactors.


beau-charman (approved):
Yeah this was weird, that file reappeared in my local feature branch on my last rebase from develop but wasn't visible in the remote develop after I merged in the PR, so I just left it.

Thank you.

PR #289, opened 2021-01-11
[Feature][AES-2264] Further testing suite improvements

beau-charman:
In line with https://aesoponline.atlassian.net/browse/AES-2264 this PR will further improve the static and unit testing efforts. 

- `.lintstagedrc` addition. 
- `jest-axe` and `jest-canvas-mock` dependancies.
- Improved test set up.
- Additions to the RodioGroup, Overlay, Modal and Select tests and coverage


src/components/TextOverFullWidthAsset/TextOverFullWidthAsset.spec.js
```diff
@@ -60,8 +52,27 @@ describe('<TextOverFullWidthAsset />', () => {
...
+  afterAll(() => {
+    global.console.error.mockRestore();
+  });
+
+  it('should not render anything and throw console error if content and mediaType are not valid', () => {
+    render(<TextOverFullWidthAsset content="" mediaType="Audio" />);
+
+    const child = screen.queryByTestId(/data-testid-TextOverFullWidthAsset/i);
+
+    expect(child).not.toBeInTheDocument();
```
MkMan, line 75: I usually do this to assert nothing is rendered, less code 😃 
```js
const { container } = render(<TextOverFullWidthAsset content="" mediaType="Audio" />);

expect(container).toBeEmptyDOMElement();
```

src/components/Video/Video.js
```diff
@@ -119,7 +119,7 @@ const Video = forwardRef(function VideoRef(
   return (
     <div className={classSet} id={id} ref={ref} role="group">
       {!!fallbackImage && (!hasVideo || hasAutoplay) && (
-        <figure className={styles.fallbackImageFigure} id={id} ref={ref}>
```
MkMan, line 122: Good pick up 👌

src/customHooks/useHasMounted/useHasMounted.spec.js
```diff
@@ -0,0 +1,21 @@
...
+      hasMounted = useHasMounted();
+
+      return null;
+    };
+
+    expect(hasMounted).toBeFalsy();
+
+    render(<ComponentWithHook />);
+
+    expect(hasMounted).toBeTruthy();
```
MkMan, line 19: I have seen seemingly weird things happen with `toBeTruthy` before. Someone once forgot to invoke a function and we were getting tests passing when they shouldn't. Rough example below.

```js
const giveMeFalse = () => false;

expect(giveMeFalse).toBeTruthy(); // will pass
expect(giveMeFalse()).toBeTruthy(); // will fail
```

Obviously the issue was because of a typo but still I've since preferred to avoid it in codebases and go for the more explicit `toBe(true)` instead.

src/components/Transition/Transition.spec.js
```diff
@@ -16,14 +16,24 @@ describe('<Transition />', () => {
...
+describe('<Transition /> required props', () => {
+  beforeAll(() => {
+    jest.spyOn(global.console, 'error').mockImplementation(() => {});
+  });
+
+  afterAll(() => {
+    global.console.error.mockRestore();
+  });
+
+  it('should return null and throw a prop error if child is not a valid element', () => {
```
MkMan, line 30: The test description is outdated, mainly the *return null* part. Can you please add an assertion for it or remove that bit from the test description 🙂

setupTests.js
```diff
@@ -1,10 +1,52 @@
 import 'regenerator-runtime/runtime';
 import '@testing-library/jest-dom/extend-expect';
+import 'jest-axe/extend-expect'; // eslint-disable-line import/no-unresolved
+/**
+ * `Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)`
+ *  https://stackoverflow.com/questions/48828759/unit-test-raises-error-because-of-getcontext-is-not-implemented
+ */
+import 'jest-canvas-mock'; // eslint-disable-line import/no-unresolved
```
kevin-ho87, line 8: what does `jest-axe` and `jest-canvas-mock` do?

src/components/DefinitionList/DefinitionList.spec.js
```diff
@@ -19,7 +19,7 @@ describe('<DefinitionList />', () => {
   it('should return `null` if the `items` prop is not a populated array', () => {
     render(<DefinitionList items={[]} />);
 
-    const child = screen.queryByTestId('data-testid-DefinitionList');
+    const child = screen.queryByTestId(/data-testid-DefinitionList/i);
```
zack76, line 22: why it's using regExp here

src/components/Heading/Heading.spec.js
```diff
@@ -10,21 +10,32 @@ describe('<Heading />', () => {
...
+    const heading = screen.getByRole('heading', { name: /test heading/i });
 
     expect(heading).toBeInTheDocument();
     expect(container).toMatchSnapshot();
   });
+});
+
+describe('<Heading /> required props', () => {
+  beforeAll(() => {
+    jest.spyOn(global.console, 'error').mockImplementation(() => {});
```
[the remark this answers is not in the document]
kevin-ho87: Does this watch for errors thrown from somewhere?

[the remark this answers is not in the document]
MkMan: Yes 🙂  It also replaces the default `console.error` implementation with an empty one (`() => {}`) so the real method doesn't get called and clutter the logs

.lintstagedrc
```diff
@@ -0,0 +1,12 @@
+{
+  "**/*.{js,json,md}": [
+    "prettier --write"
+  ],
+  "**/*.js": [
+    "eslint --ignore-path .eslintignore --config .eslintrc",
+    "jest --findRelatedTests --coverageThreshold '{}'"
```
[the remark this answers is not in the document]
beau-charman: Like I said, it's super quick. 

GEL is going to be used by are large team soon, trying to make static and unit testing as strict as reasonable

src/components/Heading/Heading.spec.js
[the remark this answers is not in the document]
beau-charman: Once of those 'it's something I see people far smarter than I do' type situations. global is safe.

setupTests.js
```diff
@@ -1,10 +1,52 @@
...
+ * and the following lines removes the warning.
+ */
+Object.defineProperty(HTMLMediaElement.prototype, 'muted', {
+  set: () => {},
+});
+
+/**
+ * Globally mock the `uuid` package
+ */
+jest.mock('uuid', () => {
```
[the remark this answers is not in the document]
beau-charman: Snapshots using a component that utilises uuid will be different each run - so by the time you mock individually, most test files will contain the mock set up. 

Hence putting it in setupTests - which should be the first place a dev looks before using a code base in regards to testing anyway so they are familiar with what is happening beyond each individual file. 

Plus, you can re-mock the implementation of a mock function

src/components/SubNav/SubNav.js
```diff
@@ -15,16 +15,22 @@ const SubNav = forwardRef(
...
     const classSet = cx(
       styles.base,
       styles[currentTheme],
       { [styles.select]: isSelect },
       className,
     );
-    const isSmallOrMediumViewport = ascertainIsSmallOrMediumOnlyViewport();
+
+    const isSmallOrMediumViewport = ascertainIsSmallOrMediumOnlyViewport(
+      count.current++,
```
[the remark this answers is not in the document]
beau-charman: neither do I :(

src/components/TextOverFullWidthAsset/TextOverFullWidthAsset.spec.js
beau-charman: I dig. I am just following the principals of 'testing what the user experiences, rather than implementation details'. 

`queryByTestId` is a bad example, but say, for a button and `queryByLabel`, i'd rather the test read as 'if there is no on click action provided, I shouldn't see a button with a label of "click me!"' - rather than 'if there is no on click action provided, render an empty div to the document object model'

Hopefully that makes sense.

src/customHooks/useHasMounted/useHasMounted.spec.js
beau-charman: "Obviously the issue was because of a typo" - this is key though, false positives and such due to a missing character can produce passing tests that shouldnt be.

beau-charman: plus hasMounted *should* return a boolean, so this is a double win with type checking

setupTests.js
beau-charman: jest axe https://github.com/nickcolley/jest-axe is simply there to help triple check interactive elements for accessibility coverage. 

jest canvas mock https://github.com/hustcc/jest-canvas-mock does just that, has jest dom doesnt have a full representation of the html element canvas

src/components/DefinitionList/DefinitionList.spec.js
beau-charman: as we are not testing anything computational here, testing for the string in any format is what we are after (what the string is exactly doesnt matter).

say if someone changed the prop to `data-testId="DATA-testid-DefinitionList"`, in this situation the test shouldnt fail, nor should it need to be updated as the test could still pass if the implementation was tested against so strictly. 

a another example: 

```
LabelComponent.js

const Label = ({ isVisible, content }) => <span style={{ display: isVisible? 'block' : 'none' }}>{content}</span>

Test.spec.js

test.('that if there is a isVisible prop, that the user can see the label')

   <Label isVisible={true} content="test-label">

   .... output <span style={{ display: block }}>test-label</span>

   getByLabel("test-label") > true
```

The above test would resolve to true, however, if oneone changed the implementation of the component, to say, render `>{content.toUpperCase()}<` the test would fail - and in this context for what we are testing for it shoudnt

src/customHooks/useHasMounted/useHasMounted.spec.js
MkMan: Hmm 🤔   I see what you're saying but `toBe(true)` is still more specific than `toBeTruthy`. If something changes and `hasMounted` starts returning an object this test will continue to pass even though it shouldn't because the desired value is `true`.

beau-charman: > Hmm 🤔 I see what you're saying but `toBe(true)` is still more specific than `toBeTruthy`. If something changes and `hasMounted` starts returning an object this test will continue to pass even though it shouldn't because the desired value is `true`.

agree, if the hook was `useOnMount`, it would make sense to have an object return value, maybe with the result of the onMount callback - but in this cause we are just listening for booleans

src/components/Modal/Modal.spec.js
```diff
@@ -5,28 +5,55 @@ import { Modal } from './Modal';
...
+      <Modal
+        copy={{ close: 'close modal copy' }}
+        isVisible={true}
+        onClose={handleOnClose}
+      >
+        modal test
+      </Modal>,
+    );
+
+    expect(screen.getByRole('dialog')).toBeInTheDocument();
```
kevin-ho87, line 57: was there a reason `getByRole` is used instead of `queryByRole` like the before `rerender`?

zack76: > was there a reason `getByRole` is used instead of `queryByRole` like the before `rerender`?

i remember `getByRole` will throw an error when the element is not found.

beau-charman: yeah, I expect to see an element with the role of `dialog`

beau-charman: same a query, but auto throws an error (so you dont even really need to run it in an ` expect()`, but just reads consistently)

beau-charman: legend: 

`get(All)By*`: when you DO expect to find the target - if not throw an error
`query(All)By*`: really only use if you expect NOT to find the target - reads more clearly this way too
`find(All)By*`: when you are asserting to both find or not find the target, but it is asynchronous - this function will keep trying for 4 seconds, then throw an error

MkMan (approved):

PR #290, opened 2021-01-11
AE-112 form components

MkMan:
#### Storybook / Stories
* Apply an order to the story groups in Storybook so smaller groups are at the top and easier to see
* Group `Checkbox`, `TextInput`, `Select` under a new *Form Elements* group

#### Checkbox
* `forwardRef` the component
* Add an `errorMessage` prop with error styling to be used for validation error messages
* Make the `id` prop optional and generate a unique id if it's not provided
* Enhance its Storybook with more knobs
* Add unit tests to cover the changes

#### Select
* `forwardRef` the component
* Remove/insert the error message instead of changing its opacity

#### TextInputV2
* **Added** a new text input component. Not sure on the name; there already exists a `TextInput` component but its styling doesn't match the ones used in Web UI, thus the need to create a new component.


src/components/Checkbox/Checkbox.js
```diff
@@ -1,51 +1,68 @@
...
+      id: idProp,
+      isEnabled,
+      name,
+      onChange,
+      theme,
+    },
+    ref,
+  ) => {
+    const baseClassSet = cx(styles.base, className);
+    const checkboxClassSet = cx(
```
zack76, line 25: the label color won't be changed if hasError?

src/components/Select/Select.js
```diff
@@ -1,83 +1,91 @@
...
-        theme={theme}
-        width={15}
-      />
-      <div className={styles.errorMessage}>{errorMessage}</div>
-    </div>
-  );
-};
+    return (
+      <div className={classSet}>
+        <label className={cx(styles.label)} htmlFor={name}>
```
kevin-ho87, line 54: Would the `id/name` have a `uuid` fallback like the `checkbox` component?

src/components/Select/Select.js
```diff
@@ -1,83 +1,91 @@
...
+        </select>
+        <Icon
+          className={styles.icon}
+          height={15}
+          name="chevron"
+          theme={theme}
+          width={15}
+        />
+        {errorMessage && (
+          <div className={styles.errorMessage}>{errorMessage}</div>
```
kevin-ho87, line 83: Are you able to update this to a `span` for accessibility

src/components/Checkbox/Checkbox.js
MkMan: image
The error message is red and the checkbox border colour changes. I felt if the label changed too it would be too "red" or too aggressive 😅

src/components/Select/Select.js
MkMan: I'll give it a crack 🙂

src/components/Select/Select.js
MkMan: The `Checkbox` had the `id` marked as required, and when I didn't provide one because I didn't need it that's when I saw the warning message. So, I made the change to prevent the component from breaking when an `id` isn't provided.

The `Select` component uses the `name` prop as the element's `id` and doesn't take an `id` prop. The `name` value is definitely required for form elements, so if one isn't provided things will not work anyway. That's why I didn't add that behaviour here 🙂

src/components/TextInputV2/TextInputV2.js
```diff
@@ -0,0 +1,125 @@
...
+            min={min}
+            name={name}
+            onChange={handleOnChange}
+            pattern={pattern}
+            ref={ref}
+            type={type}
+            value={value}
+          />
+          {label && (
+            <span
```
kevin-ho87, line 64: Would this need to be a `label` tag?

src/components/TextInputV2/TextInputV2.js
```diff
@@ -0,0 +1,125 @@
...
+                { [styles.moved]: value },
+                classes.label,
+              )}
+            >
+              {label}
+            </span>
+          )}
+        </div>
+        {errorMessage && (
+          <div className={cx(styles.errorMessage, classes.errorMessage)}>
```
kevin-ho87, line 77: Can this be a `span` for accessibility

src/components/Select/Select.js
kevin-ho87: ahh fair enough 👍

src/components/Select/Select.js
MkMan: Done 🙂

src/components/TextInputV2/TextInputV2.js
MkMan: Done 🙂

src/components/TextInputV2/TextInputV2.js
MkMan: Done 🙂

zack76 (approved):

src/components/Select/Select.module.css
```diff
@@ -90,17 +90,21 @@
...
 
 .errorMessage {
+  display: block;
   height: rem(14px);
   margin: 10px 0 0;
   color: var(--color-error);
   font-size: rem(14px);
   line-height: 1.4;
-  opacity: 0;
   transition: opacity 200ms cubic-bezier(0.215, 0.61, 0.355, 1);
```
hassanraza0054, line 99: Do we still need this opacity transition now that opacity isnt being used anymore?

MkMan: Good pick up!

MkMan: Done 🙂

kevin-ho87 (approved):

PR #298, opened 2021-01-19
[Feature][AES-2175] Add tracking functionality to ProductCommerce & ProductGridItem Components

hassanraza0054:
Added the following:

**clickTrackingFunction**: Initiate click function to push tracking data to google analytics
**trackDataProps**: Included for automation testing scripts 

Related to following Web-UI PR: https://github.com/aesop/aesop-web-ui/pull/3546

### DIFF
--- a/src/components/ProductCommerce/ProductCommerce.js
+++ b/src/components/ProductCommerce/ProductCommerce.js
@@ -31,7 +31,6 @@ const ProductCommerce = React.forwardRef(
       theme,
       isActive,
       onCtaClick,
-      trackDataProps,
     },
     ref,
   ) => {
@@ -62,7 +61,7 @@ const ProductCommerce = React.forwardRef(
     const RADIO_GROUP_DATA_TEST_REF = 'PRODUCT_COMMERCE_VARIANT_SELECT';
 
     return (
-      <div className={classSet} id={id} ref={ref} {...trackDataProps}>
+      <div className={classSet} id={id} ref={ref}>
         <Transition isActive={isActive} type={TRANSITIONS.TYPE.SLOW_FADE}>
           <SectionHeading
             className={styles.header}
@@ -171,7 +170,6 @@ ProductCommerce.propTypes = {
   size: PropTypes.string,
   isActive: PropTypes.bool,
   theme: PropTypes.oneOf(['dark', 'light']),
-  trackDataProps: PropTypes.object,
 };
 
 ProductCommerce.defaultProps = {
--- a/src/components/ProductCommerce/ProductCommerce.js
+++ b/src/components/ProductCommerce/ProductCommerce.js
@@ -30,7 +30,7 @@ const ProductCommerce = React.forwardRef(
       size,
       theme,
       isActive,
-      ctaClickFunction,
+      onCtaClick,
       trackDataProps,
     },
     ref,
@@ -132,7 +132,7 @@ const ProductCommerce = React.forwardRef(
               <Hyperlink
                 className={styles.cta}
                 isAlternate={false}
-                onClick={() => ctaClickFunction()}
+                onClick={() => onCtaClick()}
                 style={HYPERLINK_STYLE_TYPES.INTERNAL_BUTTON_LINK}
                 url={cta.url}
               >
@@ -148,7 +148,7 @@ const ProductCommerce = React.forwardRef(
 
 ProductCommerce.propTypes = {
   className: PropTypes.string,
-  ctaClickFunction: PropTypes.func,
+  onCtaClick: PropTypes.func,
   copy: PropTypes.shape({
     addToCart: PropTypes.shape({
       cartAction: PropTypes.string,
--- a/src/components/ProductCommerce/ProductCommerce.js
+++ b/src/components/ProductCommerce/ProductCommerce.js
@@ -30,6 +30,8 @@ const ProductCommerce = React.forwardRef(
       size,
       theme,
       isActive,
+      trackFunc,
+      trackDataProps,
     },
     ref,
   ) => {
@@ -60,7 +62,7 @@ const ProductCommerce = React.forwardRef(
     const RADIO_GROUP_DATA_TEST_REF = 'PRODUCT_COMMERCE_VARIANT_SELECT';
 
     return (
-      <div className={classSet} id={id} ref={ref}>
+      <div className={classSet} id={id} ref={ref} {...trackDataProps}>
         <Transition isActive={isActive} type={TRANSITIONS.TYPE.SLOW_FADE}>
           <SectionHeading
             className={styles.header}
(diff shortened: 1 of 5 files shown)


src/components/ProductGridItem/ProductGridItem.js
```diff
@@ -51,9 +64,14 @@ const ProductGridItem = React.forwardRef(
     const RADIO_GROUP_DATA_TEST_REF = 'PRODUCT_GRID_ITEM_VARIANT_SELECT';
 
     return (
-      <div className={classSet} id={id} ref={ref}>
```
MkMan, line 54: Shouldn't the ref be assigned to the "main" part of the component? Moving it may also cause issues for existing usages of `ProductGridItem` that rely on the ref being on the top level `div`.

src/components/ProductCommerce/ProductCommerce.js
```diff
@@ -60,7 +62,7 @@ const ProductCommerce = React.forwardRef(
     const RADIO_GROUP_DATA_TEST_REF = 'PRODUCT_COMMERCE_VARIANT_SELECT';
 
     return (
-      <div className={classSet} id={id} ref={ref}>
+      <div className={classSet} id={id} ref={ref} {...trackDataProps}>
```
MkMan, line 65: What's inside `trackDataProps`?

hassanraza0054: Its tracking information such as name, price and so on, that's necessary for Michelles automation test cases to hook onto the dom element and run!

src/components/ProductGridItem/ProductGridItem.js
hassanraza0054: Nice catch I agree. Thanks for that!

hassanraza0054: Done!

src/components/ProductCommerce/ProductCommerce.js
zack76: > Its tracking information such as name, price and so on, that's necessary for Michelles automation test cases to hook onto the dom element and run!

sounds like testDataProps as it's for testing?

MkMan: Ah ok 🤔  Is it a new requirement? The component has been in use without them so far

src/components/ProductCommerce/ProductCommerce.js
```diff
@@ -30,6 +30,8 @@ const ProductCommerce = React.forwardRef(
       size,
       theme,
       isActive,
+      clickTrackingFunction,
```
MkMan, line 33: I think it would be good to rename this prop to make it more generic instead of restricted to "tracking" 🙂  For example `ctaClickFunction`. Same with the `ProductGridItem`.

src/components/ProductCommerce/ProductCommerce.js
hassanraza0054: So they were already part of the encapsulating div in ProductTracker.js. I didn't see the point in that as the div only existed to contain those data tracking props. I figured the better way would be to shift them into the gel component entirely, and it made sense to me intuitively. The assumption I had is that while gel shouldn't have dependent logic inside it, keeping it mostly pure as a visual component library, it does make sense to have standardised testing properties across all gel components to keep them consistent with our test cases.

src/components/ProductCommerce/ProductCommerce.js
hassanraza0054: Done! Ty

src/components/ProductCommerce/ProductCommerce.js
beau-charman: `trackDataProps` has a prop type of object and is just spread on the html element. 

Is there any way this could be done where the implementation happens? 

Like, web ui sets up the tracking data, and web ui is the application that uses the tracking data. 

From an API point of view too, web-ui doesnt know where the tracking data props go or how they are used, and GEL just receives a random object and spreads it on a html element; there isnt even any checking to see if they are valid html attributes.

src/components/ProductCommerce/ProductCommerce.js
```diff
@@ -30,6 +30,8 @@ const ProductCommerce = React.forwardRef(
       size,
       theme,
       isActive,
+      ctaClickFunction,
```
beau-charman, line 33: also, could we keep the event handler names consistent? `onCTAClick`, for example.

src/components/ProductGridItem/ProductGridItem.js
```diff
@@ -51,9 +64,13 @@ const ProductGridItem = React.forwardRef(
...
+      <div className={classSet} id={id} {...trackDataProps} ref={ref}>
```
beau-charman, line 67: same here for the `trackDataProps`

beau-charman (approved):

src/components/ProductCommerce/ProductCommerce.js
hassanraza0054: The original implementation was just an encapsulating div with the tracking data props inside it. Should I just go about it like that, and remove trackDataProps from the gel component?

hassanraza0054: image


Correction. It's just a div with the tracking props themselves. So it can be any div, whether the gel div or another one at a higher level. I feel in this case, there's no significant pull to whether this data is in either gel or web-ui. I could create a div returned as well like the original implementation, with display as none, and remove the trackingprops from gel altogether. 

How's that?

src/components/ProductGridItem/ProductGridItem.js
hassanraza0054: Hey same response as per the above comment. What're your thoughts?

src/components/ProductCommerce/ProductCommerce.js
hassanraza0054: Done! Ty

src/components/ProductCommerce/ProductCommerce.js
beau-charman: 👍

src/components/ProductGridItem/ProductGridItem.js
hassanraza0054: Done! Added a 'none' display div in **web-ui** and attached the **trackDataProps** to it

src/components/ProductCommerce/ProductCommerce.js
hassanraza0054: Done!

MkMan (approved):

MkMan:
The test coverage is [0.01% less than the minimum required](https://github.com/aesop/aesop-gel/pull/298/checks?check_run_id=1746035298#step:5:532) and is failing the build 😂  🤦‍♂️

hassanraza0054:
> The test coverage is [0.01% less than the minimum required](https://github.com/aesop/aesop-gel/pull/298/checks?check_run_id=1746035298#step:5:532) and is failing the build 😂 🤦‍♂️

Just made the change!

PR #300, opened 2021-01-20
AE-113 Create DynamicForm

MkMan:
This PR is bigger than it looks. Most of it is fixtures/tests/package-lock/"duplicated" code so come on in and don't be put off 😄 

This PR adds a component that accepts a form structure as a JSON and renders a form. It uses a library called [React Hook Form](https://react-hook-form.com/) under the hood for form management.

This is a rough breakdown of what's happening:
- There is a list of accepted form element types [here](https://github.com/aesop/aesop-gel/pull/300/files#diff-592a324defde197e0c562ef275d5d3d417177ac1ad8ce778a5512f2dd6a4500c)
- For each one of the accepted types above there is a wrapper component. Its job is hook the form to the Gel component and pass it the props it needs. The wrappers are defined [here](https://github.com/aesop/aesop-gel/pull/300/files#diff-139cda0244571db70dfd00aee61c50a809e0a36c4485cf31f430f010bcbe7f09). They look fairly similar to one another but I preferred this way instead of 1 wrapper where the props get mixed up and components get props they don't need.
- The validators are [here](https://github.com/aesop/aesop-gel/pull/300/files#diff-6c5ab0879789f933afcc4bfca83da17be81909e78a87bb440e6cbb06f7899fa2)
- The DynamicForm component loops through the schema to render the form elements it specified by using the wrappers and validators.

I also put in a bit of readme/guide in the component's Storybook page

**Screenshot from Storybook**
image


src/compositions/DynamicForm/DynamicForm.module.css
```diff
@@ -0,0 +1,56 @@
+@import 'styles/index';
+
+.formRowWrapper {
+  display: flex;
+  flex-direction: column;
```
kevin-ho87, line 5: Is flex needed on mobile?

src/compositions/DynamicForm/wrappers/Checkbox.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable react/prop-types */
+import React from 'react';
+import { Controller } from 'react-hook-form';
+import { Checkbox } from '~/components/Checkbox';
+
+export const CheckboxWrapper = ({
```
kevin-ho87, line 6: Would `forwardRef` be needed to pass through `ref` ?

src/compositions/DynamicForm/DynamicForm.stories.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+#### Maximum length
+
+This rule is only applicaple to the **`TextField`** field type. It will show an error message if the value in the field exceeds the length
+specified by the rule. It is defined as follows
+
+```json
+"maxLength": {
+  "value": 30,
+  "message": "Error message to show when more than 30 characters are enrtered",
```
kevin-ho87, line 96: typo `enrtered`

src/compositions/DynamicForm/DynamicForm.stories.mdx
```diff
@@ -0,0 +1,110 @@
...
+    <DynamicForm
+      formSchema={knobs.select('formSchema', {
+        'None': null,
+        'Au Sign up form': DynamicFormFixture.auSignUp,
+        'Korean Sign up form': DynamicFormFixture.krSignUp,
+      }, DynamicFormFixture.auSignUp, 'Content')}
+      onSubmit={DynamicFormFixture.onSubmit}
+      theme={knobs.select('theme', ['dark', 'light'], 'dark', 'Presentation')}
+    >
+      <Button theme="dark" type="submit">
```
kevin-ho87, line 23: I wonder if we need a button row wrapper just for spacing. Or should this be handled in web-ui?

src/compositions/DynamicForm/DynamicForm.js
```diff
@@ -0,0 +1,137 @@
...
+                  options,
+                  styling,
+                  subtype,
+                  testRef,
+                  type,
+                  validation,
+                },
+                index,
+              ) => {
+                const InputField = componentMap[type];
```
hassanraza0054, line 56: Could you explain how type and subtype are being used here? I understand that subtype is basically type for the textfield, and the below returned input field is an HOC that becomes a specific kind of input field based on its type prop.

The only uniqueness of subtype atm is that it becomes **type** for **textField**, with options such as "password" and later possibly "username" or "email". This is in turn is used in the TextInput field. Am I correct in this understanding?

MkMan: Spot on 👌 
For now it's only relevant for text fields but in the future say the dropdown component comes in different types, it could be used for it too.

src/compositions/DynamicForm/wrappers/Checkbox.js
MkMan: From `DynamicForm` to `CheckboxWrapper`?

src/compositions/DynamicForm/DynamicForm.stories.mdx
MkMan: I don't fully understand what you mean sorry 🙏  
I felt it's up to the parent component/consuming app to provide whatever else they want to appear after the form and style it however they wish.

src/compositions/DynamicForm/wrappers/Checkbox.js
kevin-ho87: Yeah, but it may not be needed. Just asking as I did notice some components having `forwardRef` in them

MkMan: Cool I understand what you mean now 👍 
At this stage I didn't see the need for the form component to have individual refs to each of its children form elements. If in the future we find that's needed we can look at adding it in

src/compositions/DynamicForm/DynamicForm.module.css
MkMan: Good pick up, managed to make do without it. Thanks 🙂

hassanraza0054 (approved):

src/compositions/DynamicForm/DynamicForm.js
```diff
@@ -0,0 +1,137 @@
...
+                  <InputField
+                    className={formFieldClassSet}
+                    control={control}
+                    dataTestRef={testRef}
+                    defaultValue={
+                      defaultValues?.[name] || defaultValueFromSchema
+                    }
+                    errorMessage={errors[name]?.message}
+                    id={id}
+                    key={index}
```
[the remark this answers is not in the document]
MkMan: That's just for React optimisation 🙂   You can find more details about it [here](https://reactjs.org/docs/lists-and-keys.html#keys).

zack76 (approved):

src/compositions/DynamicForm/DynamicForm.spec.js
```diff
@@ -0,0 +1,152 @@
...
+    expect(inputField).toHaveValue('Clark');
+  });
+
+  it('should render all fields that are registered in the map', () => {
+    const formSchema = [
+      [selectFieldSchema, textFieldSchema],
+      [checkboxFieldSchema],
+    ];
+
+    const { getByRole } = render(
```
beau-charman, line 103: I might have missed a bunch of these, but can we move away from destructing the query method from the render function, and use `screen` from '@testing-library/react'? Cheers!

src/compositions/DynamicForm/DynamicForm.stories.mdx
beau-charman: Yeah, avoid having visual helpers in storybook, it gives an incorrect representation of how the components looks and behave together. For example, I have only just noticed a storybook only class of `helper` wrapped around this component https://aesop.github.io/aesop-gel/?path=/story/components-productgriditem--base-component which falsely represents how the component will align

beau-charman (approved):
🎉

beau-charman:
Could you just replace the destructured test methods with `screen.` calls, and then this is great!

src/compositions/DynamicForm/DynamicForm.spec.js
MkMan: Sure 😃 

Stack Overflow question discussing [this](https://stackoverflow.com/a/61651524/8459521) for anyone interested.

MkMan: Done :)

PR #304, opened 2021-01-29
Improve GH action

MkMan:
This stops running `build:dev` for release as `build:prod` runs after it anyway

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,4 +1,4 @@
-name: On Branch Push
+name: Prepare Release
 on:
   push:
     branches:
@@ -7,19 +7,14 @@ on:
       - '**'
 
 jobs:
-  test_and_release:
-    name: Test & Release
+  prepare_release:
+    name: Prepare Release
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     steps:
       - uses: actions/checkout@v2
         with:
           persist-credentials: false
-
-      - name: Create variables
-        run: |
-          echo "BRANCH=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-          echo "RELEASE_BRANCHES=['develop', 'master']" >> $GITHUB_ENV
 
       - name: Install
         run: npm ci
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yml
```diff
@@ -26,9 +31,11 @@ jobs:
         run: npm run test
 
       - name: Build for Release
+        if: "contains(env.RELEASE_BRANCHES, env.BRANCH)"
```
MkMan, line 34: @maxlew Can you think of a way to avoid duplicating this condition across two steps? I tried created a boolean env variable but couldn't get it right.

maxlew: The second `npx semantic-release` command should be fine to run on all branches, if there's no config for that branch set up in here - https://github.com/aesop/aesop-gel/blob/develop/.releaserc it wont do anything. 

I'd still run `build:prod` on all branches too, it'll cause the job to be slower but will pick up bugs that tests and linting wont.

MkMan: > I'd still run build:prod on all branches too, it'll cause the job to be slower but will pick up bugs that tests and linting wont.

Riiight, so that's what it was there for 💡 
The reason I made this change was to stop running `build:dev && build prod` because they constituted a big part of the build time. Seeing that they're needed then I'll drop these changes.

maxlew: I reckon it's not worth running them both, not sure why that was done

package.json
```diff
@@ -19,7 +19,7 @@
     "generate-component": "./scripts/componentBoilerplate/create.sh $element",
     "lint": "npm run eslint && npm run stylelint",
     "prepublishOnly": "npm run release",
-    "release": "npm run build:dev && npm run build:prod",
+    "release": "npm run build:prod",
```
MkMan, line 22: What's the reason behind running both `build:dev` and `build:prod` , instead of just `build:prod` @maxlew @beau-charman?

MkMan: https://github.com/aesop/aesop-gel/pull/304#discussion_r566573397

maxlew (approved):

PR #307, opened 2021-02-01
chore(storybook): deploy storybook task

beau-charman:
I actually took some time to finally set up a deploy task for gh-pages: https://aesop.github.io/aesop-gel/. so no more manual copy and pasting 🥳


zack76 (approved):

package.json
```diff
@@ -14,6 +14,7 @@
     "build:dev": "NODE_ENV=development rollup -c",
     "build:prod": "NODE_ENV=production rollup -c",
     "build:watch": "NODE_ENV=production rollup -c -w",
+    "deploy-storybook": "gh-pages -d storybook-static",
```
MkMan, line 17: Should we add this to our GH workflow to run when release runs?

beau-charman: I was going to ask you for a hand writing an action for this, just had no time. but yeah, that would be great

MkMan: Sure 🙂  
I am just finishing up something at the moment, once done I'll have a look.
I know there is a semantic-release plugin that's used to execute commands as part of the release process. I was going to look into that first.

MkMan (approved):

hassanraza0054 (approved):

PR #308, opened 2021-02-02
[Update] Add To Cart Button Memory Leak

beau-charman:
Refactor and test for Add To Cart Button timeout memory leak on unmount


src/components/AddToCartButton/AddToCartButton.js
```diff
@@ -123,7 +114,11 @@ const AddToCartButton = ({
           shouldShowUpdateSuccessMessage ? `${dataTestRef}_SUCCESS` : undefined
         }
       >
-        <span>{updateNotificationLabel}</span>
+        {shouldShowUpdateSuccessMessage ? (
+          <span>{updateNotificationLabel}</span>
+        ) : (
+          <span> </span>
+        )}
```
MkMan, line 121: Is an empty `<span>` needed? Would `{shouldShowUpdateSuccessMessage && <span>{updateNotificationLabel}</span>}` achieve the same desired outcome?

src/components/AddToCartButton/AddToCartButton.spec.js
```diff
@@ -38,6 +38,23 @@ describe('<AddToCartButton />', () => {
...
+  afterAll(() => {
+    global.console.error.mockRestore(); // eslint-disable-line no-console
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
```
MkMan, line 54: Does this need to run before and after each test? I thought just before would be enough 🤔

src/components/AddToCartButton/AddToCartButton.spec.js
```diff
@@ -157,4 +174,33 @@ describe('<AddToCartButton />', () => {
 
     expect(loading).toBeInTheDocument();
   });
+
+  it('should have no active timers on unmount (should not throw state update error)', async () => {
```
MkMan, line 178: What do you think about the arrange-act-assert pattern? I find it to be a good way on organising long tests.

MkMan:
> Refactor and test for Add To Cart Button timeout memory leak on unmount

I am not very familiar with this side of things. What was causing the memory leak? :)

src/components/AddToCartButton/AddToCartButton.js
beau-charman: For the way the original (web-ui) component was written, the button always had two lines of text. When writing the test for the timeout function on click, the assertion is that after the click the user should see the `added to cart text`; but in reality the user could kinda always 'see' it. So, in following the principals of test driven development, I updated the component to correctly reflect the test

beau-charman:
> > Refactor and test for Add To Cart Button timeout memory leak on unmount
> 
> I am not very familiar with this side of things. What was causing the memory leak? 🙂

The set timeout still runs after the component unmount, which will through a reference error

src/components/AddToCartButton/AddToCartButton.spec.js
beau-charman: good pick up

src/components/AddToCartButton/AddToCartButton.spec.js
beau-charman: agree, but in terms of user flow, if something fails and can short circuit the test, it kinda makes more sense. and this one in particular, after the render, is only 8 lines.

src/components/AddToCartButton/AddToCartButton.js
MkMan: Ah ok, I was not aware of what the component looked like before 👍   I am not 100% sure on how it needs to look/behave but my opinion is that if we don't need the `span` when it doesn't have content we should update the component and the test. In the same way we won't break a component because an existing test was asserting wrong behaviour 😄

beau-charman:
@MkMan , for example, if I comment out lines 16 - 18 in `src/components/AddToCartButton/AddToCartButton.utils.js`, I will get this error:

MkMan:
> @MkMan , for example, if I comment out lines 16 - 18 in `src/components/AddToCartButton/AddToCartButton.utils.js`, I will get this error:
> 
> Screen Shot 2021-02-03 at 8 19 46 am
> 
> As there is still an async callstack function trying to run
> 
> ```
>   timeout.current = setTimeout(() => {
>   👉  setIsDelayedDisabled(false);
>   }, 2500);
> ```

Awesome! Thanks for explaining 😃

src/components/AddToCartButton/AddToCartButton.utils.js
```diff
@@ -0,0 +1,24 @@
...
+      setIsDelayedDisabled(true);
+
+      timeout.current = setTimeout(() => {
+        setIsDelayedDisabled(false);
+      }, 2500);
+    }
+
+    return function cleanup() {
+      clearTimeout(timeout.current);
+    };
```
kevin-ho87, line 18: 👍  Ah I didn't know about adding clearTimeout in the cleanup function. Learned something new

MkMan: Same 🙌  😃

kevin-ho87 (approved):

src/components/AddToCartButton/AddToCartButton.js
beau-charman: The span will force spacing in regards to the slide animation. A method ported over from web UI.

MkMan: I see, it's needed for functionality. Makes sense then 🙂

MkMan (approved):

PR #309, opened 2021-02-05
fix: update the pattern and order of Content hub articles

zack76:
Update the order of Content hub articles

**4 articles:**
Screen Shot 2021-02-05 at 11 49 42

**3 articles:**
Screen Shot 2021-02-05 at 11 49 49

**2 articles:**
Screen Shot 2021-02-05 at 11 49 54

**1 article:**
Screen Shot 2021-02-05 at 11 50 00


src/components/ContentHubArticleList/ContentHubArticleList.js
```diff
@@ -39,89 +39,94 @@ const ContentHubArticleList = ({
...
               />
             )}
           </div>
           <div className={bottomLeftClassSet}>
             {count >= 1 && (
               <ContentHubArticle
-                articleRef={(articles[count - 2] || articles[0]).articleRef}
-                category={(articles[count - 2] || articles[0]).category}
-                dataTestRef={(articles[count - 2] || articles[0]).id}
+                articleRef={articles[count === 2 ? 0 : count - 1].articleRef}
```
MkMan, line 93: Is it possible to make these into a variable/function instead of duplicating them? It'll have the added benefit of explaining what they mean through the name of the variable/function 🙂 

Same for `count === 2 ? 1 : articleNo` if possible

zack76: > Is it possible to make these into a variable/function instead of duplicating them? It'll have the added benefit of explaining what they mean through the name of the variable/function 🙂
> 
> Same for `count === 2 ? 1 : articleNo` if possible

Yeah, it's actually for the order when the number of articles is less than 3 the order of the pattern will be clockwise, otherwise it'll be counterclockwise. It's quite tricky because the order of article is not actually same as the order of its place. And i can barely find the common thing to make it a function

src/components/ContentHubArticleList/ContentHubArticleList.module.css
```diff
@@ -52,6 +52,7 @@
 }
 
 .nonFirstRow {
+  display: none;
   grid-row: 1 / 3;
```
kevin-ho87, line 56: is this line needed if it is hidden?

zack76: > is this line needed if it is hidden?

Yes. Otherwise, it will leave an empty space

src/components/ContentHubArticleList/ContentHubArticleList.js
```diff
@@ -45,83 +49,91 @@ const ContentHubArticleList = ({
...
           <div className={topLeftClassSet}>
-            {count >= 3 && (
+            {count >= (articleIndex = pattern === 0 ? 4 : 3) && (
               <ContentHubArticle
-                articleRef={articles[0].articleRef}
-                category={articles[0].category}
-                dataTestRef={articles[0].id}
-                horizontalThumbnail={articles[0].horizontalThumbnail}
-                id={articles[0].id}
+                articleRef={articles[count - articleIndex].articleRef}
```
kevin-ho87, line 54: would it work if this is stored in a variable eg:
```
const theArticle = articles[count - articleIndex]
```
then use `theArticle. articleRef`

src/components/ContentHubArticleList/ContentHubArticleList.js
```diff
@@ -29,100 +29,114 @@ const ContentHubArticleList = ({
...
     styles[`article`],
+    {
+      [styles.nonFirstRow]: count === 1 && pattern === 1,
+    },
   );
   const leftMiddleArticleClassSet = cx(styles.leftMiddleArticle);
   const middleArticleClassSet = cx(styles.middleArticle);
   const rightMiddleArticleClassSet = cx(styles.rightMiddleArticle);
 
+  const getArticleIndex = (isOnTop, isOnLeft) => {
```
zack76, line 46: @kevin-ho87 update by creating these two functions

zack76: > @kevin-ho87 update by creating these two functions

updated by line 46 - 58

src/components/ContentHubArticleList/ContentHubArticleList.js
zack76: > would it work if this is stored in a variable eg:
> 
> ```
> const theArticle = articles[count - articleIndex]
> ```
> 
> then use `theArticle. articleRef`

updated by line 46 - 58

MkMan (approved):

PR #314, opened 2021-02-09
[Feature][AES-2467] Video Subtitle / Captions

beau-charman:
This PR address the GEL partition of https://aesoponline.atlassian.net/browse/AES-2467 by adding a captions track, toggle button, props, logic and styling. 

https://aesop.github.io/aesop-gel/?path=/story/components-video--video-with-captions-track

~As a by-product of testing this, I also found that the `uuid` package was rendering new ids on every render (bad for a component like Video where the progress causes re-renders), so this has been replaced with https://github.com/thearnica/react-uid#readme 🥳  - snapshots updated to accommodate this.~

The `should show loading html when clicked` test in AddToCartButton needs to be re written as I was getting false negatives in local testing due to the async nature of the test.

The new storybook also takes into account default function properties now when generating docs, so I am starting the process of removing `defaultProps` from components, as it is too verbose and too much implementation detail. 

~Lastly, the inconsistency of `cleanUp` and `cleanup` needed to be addressed 😓 .~


beau-charman:
@MkMan , will React UID affect the use of IDs in checkbox and inputv2?

src/components/Video/components/Controls/Controls.js
```diff
@@ -138,6 +139,20 @@ const Controls = ({
...
+        <Button
+          className={cx(styles.captionsToggle, {
+            [styles.hidden]: !isActiveVideoControlActive && isPlaying,
+            [styles.isActive]: captions?.isActive,
+          })}
+          isInline={true}
+          onClick={handleOnCaptionsToggleClick}
+          title={captions?.toggleButtonTitle}
+        >
+          CC
```
zack76, line 152: should the caption button be using an icon or SVG rather than text "CC"?

MkMan:
Is it possible to split this into smaller PRs 🙏 ? 54 files is a lot of files to review 😄 
It's better to keep refactors/enhancements/chores separate from feature work because it makes reverting easier if needed

MkMan:
> @MkMan , will React UID affect the use of IDs in checkbox and inputv2?

It should be fine if it offers the same functionality as the old library. They were used as a fallback `id` (to link labels to inputs) in case one isn't provided.

src/components/Checkbox/Checkbox.js
```diff
@@ -21,6 +21,7 @@ const Checkbox = forwardRef(
     },
     ref,
   ) => {
+    const uid = useUID();
```
MkMan, line 24: Does it need to be declared as a const or can it be invoked if needed like before?

src/components/Video/components/Controls/Controls.js
beau-charman: CC

src/components/Checkbox/Checkbox.js
beau-charman: happy to invoke inline

src/components/Video/components/Controls/Controls.js
beau-charman: CC

> Is it possible to split this into smaller PRs 🙏 ? 54 files is a lot of files to review 😄
> It's better to keep refactors/enhancements/chores separate from feature work because it makes reverting easier if needed

As illustrated in the description, the other changes were made in response to the feature PR being blocked due to false fail tests and over re-rendering optimisation 

The cleanup change was a selfish one for my mental health.

beau-charman:
> Is it possible to split this into smaller PRs 🙏 ? 54 files is a lot of files to review 😄
> It's better to keep refactors/enhancements/chores separate from feature work because it makes reverting easier if needed

As illustrated in the description, the other changes were made in response to the feature PR being blocked due to false fail tests and over re-rendering optimisation

The cleanup change was a selfish one for my mental health.

src/components/Video/components/VideoPlayer/VideoPlayer.utils.js
```diff
@@ -0,0 +1,19 @@
+import {
+  ascertainIsSmallOnlyViewport,
+  ascertainIsMediumOnlyViewport,
+  ascertainIsLargeViewport,
+} from '~/utils/viewports';
+
+const getVideoSrc = sizes => {
```
MkMan, line 7: Can you please add unit tests for this 🙏 ?

src/components/Video/components/VideoPlayer/VideoPlayer.utils.js
```diff
@@ -0,0 +1,19 @@
...
+  const { small, medium, large } = sizes || {};
```
MkMan, line 8: Wanted your thoughts on this one. Is this the same as providing a default value for `sizes`?
```js
const getVideoSrc = (sizes = {}) => {
```

src/components/Video/Video.module.css
```diff
@@ -8,6 +8,39 @@
   margin: 0;
   object-fit: cover;
 
+  video::cue {
```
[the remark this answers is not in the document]
beau-charman: Still working on the styles with Max, I might make this PR a draft

src/components/Video/components/VideoPlayer/VideoPlayer.utils.js
```diff
@@ -0,0 +1,19 @@
...
+
+  if (small && ascertainIsSmallOnlyViewport()) return small;
+
+  if (medium && ascertainIsMediumOnlyViewport()) return medium;
+
+  if (large && ascertainIsLargeViewport()) return large;
+
+  return small || medium || large || '';
```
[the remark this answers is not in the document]
beau-charman: No, like, if there is only a small provided, and the user is on a large screen, there should still be a video, that as a requirement, but `video` doesnt support responsive sources like picture does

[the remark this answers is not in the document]
MkMan: Ahhhhh, got it, makes sense 👌

src/components/Video/components/VideoPlayer/VideoPlayer.utils.js
beau-charman: Reason I am staying away from `lazy` defaults is to be explicit. In the component, one could still rely on `sizes` being falsy if not passes at all (like the typical `if (!sizes) return null`) but if we give it a "blank slate" default, we are kinda incorrectly representing what props where passed.

beau-charman: I even don't like `onClick = () =>` as a default, as one is simply declaring a memory pointer to a function that does nothing. I'd rather be explicit, and if the prop is not required, simply run a conditional check before it: 

 ```
 if (props.onClick) prop.onClick(); 
 ```

src/components/Video/Video.spec.js
```diff
@@ -8,6 +8,10 @@ describe('<Video />', () => {
     expect(Video).toBeDefined();
   });
 
+  it('should not crash application when rendered with no props', () => {
```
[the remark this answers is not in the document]
beau-charman: Just trying out a worst case scenario - trying to think of a way of stress testing all the component without manually testing without certain props. if you have any suggestions, that would be great. (obviously components with required props will throw their own errors). Maybe another item for a call discussion.

src/components/Video/components/VideoPlayer/VideoPlayer.utils.js
MkMan: That's an interesting point 🤔 
Do you know which results in more js to send to the client? That could also be a factor to consider. I think RAM nowadays is plenty

beau-charman:
```
{
  onclick = () => {},
}
...
```
onclick()
```
{
  onClick,
}
...
if (onClick) onClick();
```

and, `onClick` most likely will be turned into `g` or something anyway.

```
const f = ({g}) => {if (g) g();}
```

```
const f = ({g = () => }) => {g();}
```

MkMan: Fair enough 😃 
I'll leave this unresolved so the discussion is more visible for other reviewers

beau-charman: oooo. the js minifier turns `if (onClick) onClick();` into `g&&g()`, thats cool

src/components/Video/Video.js
```diff
@@ -12,80 +11,65 @@ import { Image } from '~/components/Image';
...
-    const videoRefCurrent = videoRef.current;
-
-    if (isScrollBasedVideo && videoRefCurrent) {
-      // eslint-disable-next-line no-new
-      new VideoScroller({ el: videoRefCurrent });
-    }
-  }, [isScrollBasedVideo]);
-
-  const pauseVideo = () => {
+  function pauseVideo() {
```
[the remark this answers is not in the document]
beau-charman: I try to bury implementation details. function block underneath the important code is a step before importing them from a seperate file. I might even declare them after the component and pass in the state setter callbacks and refs now that I look at it. also, I also don’t care about `this` or implicit returns here, so no need for an arrow function. Hiding the implementation is more important that slightly less characters and 'trendy' code in this case. Hopefully that makes sense.

src/components/Video/Video.module.css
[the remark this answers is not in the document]
beau-charman: Also, the idea was to get the style right for browser that accommodate `::cue` and other track selectors, and let other browsers use their native styles if they dont support them

src/components/Video/Video.js
```diff
@@ -12,80 +11,65 @@ import { Image } from '~/components/Image';
...
 import { Poster } from './components/Poster';
 import { VideoPlayer } from './components/VideoPlayer';
+import { useVideoScroller, useProgress } from './Video.customHooks';
 import styles from './Video.module.css';
 
 const Video = forwardRef(function VideoRef(
   {
+    captions,
     className,
     copy,
```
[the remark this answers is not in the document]
beau-charman: I think "Should the copy prop be marked as required then to alert us if it's not provided?" is more of a business logic item to as POs

zack76 (approved):

src/components/Video/components/VideoPlayer/VideoPlayer.utils.js
beau-charman: Have added an issue to chase this up 😓

PR #316, opened 2021-02-11
Chore: Replace UUID with React UID for persistent IDs

beau-charman:
https://github.com/thearnica/react-uid#readme

The uuid package was rendering new ids on every render (bad for a component like Video where the progress causes re-renders), so this has been replaced with React UID.

Snapshots updated to accommodate this.

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -1,37 +1,37 @@
 import React, { forwardRef } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import { v4 as uuidv4 } from 'uuid';
+import { useUID } from 'react-uid';
 import styles from './Checkbox.module.css';
 
 const Checkbox = forwardRef(
   (
     {
-      className,
       checkboxClassName,
+      className,
       content,
       contentClassName,
       dataTestRef,
       errorMessage,
       id: idProp,
-      isEnabled,
+      isEnabled = true,
       name,
       onChange,
-      theme,
+      theme = 'dark',
     },
     ref,
   ) => {
+    const uid = useUID();
+    const inputId = idProp || uid;
+    const errorMessageId = `${inputId}-error-message`;
     const baseClassSet = cx(styles.base, className);
+    const contentClassSet = cx(styles.content, styles[theme], contentClassName);
     const checkboxClassSet = cx(
       styles.checkBox,
       styles[theme],
       { [styles.hasError]: errorMessage },
       checkboxClassName,
     );
-    const contentClassSet = cx(styles.content, styles[theme], contentClassName);
-
-    const inputId = idProp || uuidv4();
-    const errorMessageId = `${inputId}-error-message`;
 
     return (
       <div>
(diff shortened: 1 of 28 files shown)


src/components/Checkbox/Checkbox.js
```diff
@@ -74,17 +74,4 @@ Checkbox.propTypes = {
   theme: PropTypes.oneOf(['dark', 'light']),
 };
 
-Checkbox.defaultProps = {
```
MkMan, line 77: Wouldn't this remove them from the [documentation section in Storybook](https://aesop.github.io/aesop-gel/?path=/docs/form-elements-checkbox--base-component#api)?

src/components/TextInputV2/TextInputV2.js
```diff
@@ -102,7 +103,7 @@ TextInputV2.propTypes = {
   dataTestRef: PropTypes.string,
   errorMessage: PropTypes.string,
   id: PropTypes.string,
-  isDisabled: PropTypes.bool,
+  isDisabled: PropTypes.bool /* @TODO can we please make this isEnabled */,
```
MkMan, line 106: What's the reason for preferring `isEnabled` over `isDisabled`? Personally, I assume that any component is enabled unless I explicitly disable it. `isDisabled` also maps nicer with the `disabled` attribute.

MkMan (approved):

src/components/Checkbox/Checkbox.js
beau-charman: No, another piece of engineering work I want to pitch is starting to remove `defaultProps`, as mentioned in https://github.com/aesop/aesop-gel/pull/314 where this change originated. 


> The new storybook also takes into account default function properties now when generating docs, so I am starting the process of removing `defaultProps` from components, as it is too verbose and too much implementation detail.

src/components/TextInputV2/TextInputV2.js
beau-charman: Happy to have a call about this, it's hard to explain here, but basically the API should be consistent. I have explained this a few times in past PRs. Keeping the API as 'truthy', 'positive', 'shouldBe', 'has', 'is' but default means there are no ambiguities around what `!` would do to a variable. 

`hasContent && isActive && !isDisabeld` for me is much harder to comprehend quickly than `hasContent && isActive && isEnabled` 

`disabled` is a HTML attribute - and HTML is old, however we are writing JavaScript and have the power to write better APIs, can we please just be consistent with our API. I would argue one should not follow a reactively build HTML API that was created decades ago.

src/components/Checkbox/Checkbox.js
MkMan: That's AWESOME!

src/components/TextInputV2/TextInputV2.js
beau-charman: "Affirmative", thats the word i was looking for :)

src/components/Checkbox/Checkbox.js
beau-charman: Hey, it was your updating storybook work that brought it to us!

src/components/TextInputV2/TextInputV2.js
MkMan: > but basically the API should be consistent

Definitely agree 👍 

> Happy to have a call about this

That would be great 🙂  I think I get what you mean about the affirmative but I am keen to know for certain

PR #318, opened 2021-02-15
[Chore] Refactor build tasks and normalise task names

beau-charman:
At the moment, we have no need for a 'production' build of GEL (consuming libraries / projects will deal with that.)

Furthermore, I wanted to normalise the task format, and the `-` character is one less key stroke than `:`, plus it stands out a little better when scanning code.


beau-charman:
For the error in https://github.com/aesop/aesop-gel/pull/318/checks?check_run_id=1900592740#step:7:8, I looked in the github/workflows and couldn't see this to update, where is this workflow defined?

package.json
```diff
@@ -9,24 +9,22 @@
     "dist"
   ],
   "scripts": {
-    "build": "npm run build:dev",
+    "build": "NODE_ENV=development rollup -c",
```
MkMan, line 12: Do you know if we'd still need [this](https://github.com/aesop/aesop-gel/blob/develop/rollup.config.js#L15)?

MkMan (approved):

package.json
beau-charman: Yeah, I kinda wanna get rid off it and just evaluate the ternaries that need `NODE_ENV` as the develop version

kevin-ho87 (approved):

PR #320, opened 2021-02-20
[Update] Use the IE fallback hook for React Hook Form

beau-charman:
This PR aims to address IE11 legacy issue by implementing the developer suggested code in https://react-hook-form.com/faqs#BrowserSupport, which is to use the IE hook. 

Web UI Branch using these changes can be found [here](https://github.com/aesop/aesop-web-ui/compare/feature/demo-IE-dynamicForm) and is deployed [here](https://demo-ie-dynamicform.acceptance.aesop.com/au).

Screenshot
IE

### DIFF
--- a/src/compositions/DynamicForm/DynamicForm.js
+++ b/src/compositions/DynamicForm/DynamicForm.js
@@ -1,7 +1,7 @@
 import React, { forwardRef } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import { useForm } from 'react-hook-form';
+import { useForm } from 'react-hook-form/dist/index.ie11';
 
 import { useThemeContext } from '~/contexts';
 import { componentMap } from './component-map';
--- a/src/compositions/DynamicForm/wrappers/Checkbox.js
+++ b/src/compositions/DynamicForm/wrappers/Checkbox.js
@@ -1,6 +1,6 @@
 /* eslint-disable react/prop-types */
 import React from 'react';
-import { Controller } from 'react-hook-form';
+import { Controller } from 'react-hook-form/dist/index.ie11';
 import { Checkbox } from '~/components/Checkbox';
 
 export const CheckboxWrapper = ({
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -76,6 +76,7 @@
     "@babel/plugin-transform-classes": "^7.10.4",
     "@babel/preset-env": "^7.8.4",
     "@babel/preset-react": "^7.8.3",
+    "@babel/runtime-corejs3": "^7.12.18",
```
MkMan, line 79: I am not very familiar with babel, do we need to do something to use this library?

MkMan (approved):
I hate IE :rage4: 💔

zack76 (approved):

MkMan:
Are you going to revert #321 as part of this fix?

package.json
beau-charman: Yeah same, not sure if we will need it with the combination of pollyfills, just added it while I was working on the fix

beau-charman:
> Are you going to revert #321 as part of this fix?

That is the idea, although I am not sure if I will have time to continue on with this fix, I might have to pass it on

MkMan:
> > Are you going to revert #321 as part of this fix?
> 
> That is the idea, although I am not sure if I will have time to continue on with this fix, I might have to pass it on

Happy to take this off your hands 👍

hassanraza0054 (approved):

beau-charman:
Are we forcing a performance/feature/quality-of-life hit to all user just because of IE?

MkMan:
> Are we forcing a performance/feature/quality-of-life hit to all users just because of IE?

Yep 😢 

Sadly, we can't bundle both and use the "correct" one during runtime as both would be shipped to the browser 
https://github.com/react-hook-form/react-hook-form/discussions/2603#discussioncomment-51350

beau-charman:
> > Are we forcing a performance/feature/quality-of-life hit to all users just because of IE?
> 
> Yep 😢
> 
> Sadly, we can't bundle both and use the "correct" one during runtime as both would be shipped to the browser
> [react-hook-form/react-hook-form#2603 (reply in thread)](https://github.com/react-hook-form/react-hook-form/discussions/2603#discussioncomment-51350)

yeah. absolutely fair

PR #322, opened 2021-02-22
TypeScript Migration

beau-charman:
This PR is a work in progress to migrate GEL to Typescript - Collaboration more than welcome. 

https://gel-typescript.acceptance.aesop.com/au

### Development tasks remaining

- [ ] Convert remaining `.js` components, hooks and utils
- [ ] Configure prettier correctly 
- [ ] Configure eslint to lint both `js` and `ts|x` files, but correctly override `ts|x` with `@typescript` rules
- [x] Address extension and module resolution issues in Jest and Rollup
Screen Shot 2021-02-23 at 9 21 37 am
Screen Shot 2021-02-23 at 8 45 05 am


package.json
```diff
@@ -101,7 +102,15 @@
     "@testing-library/react": "^11.2.2",
     "@testing-library/react-hooks": "^3.4.2",
     "@testing-library/user-event": "^12.5.0",
-    "@types/jest": "^25.1.1",
+    "@types/classnames": "^2.2.11",
+    "@types/jest": "^25.2.3",
+    "@types/jest-axe": "^3.5.1",
+    "@types/lodash": "^4.14.168",
+    "@types/react": "^17.0.2",
```
MkMan, line 109: Do you know if we should match the version of the types with our `react` version?

MkMan (commented):
Just curious why we're moving to the `.tsx` extension instead of just `.ts` like we were using `.js` before?

**EDIT** 5/3/21: Apparently the TS team is very much against this https://github.com/microsoft/TypeScript/issues/30503 and https://github.com/microsoft/TypeScript/issues/26489 and it's not supported.

src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
MkMan, line 1: When do we use `import React from 'react';` and `import * as React from 'react';`

package.json
beau-charman: Definitely. This PR needs a lot of work, and please if you have time, contribute these changes, or add them to the "Development tasks remaining". Thank you!

beau-charman:
> Just curious why we're moving to the `.tsx` extension instead of just `.ts` like we were using `.js` before?

Not 'moving to' as much as following the principals of typescript. TypeScript is very different to Babel, and by default it knows to apply React typings to `<JSX />` only when the `x` is present in the file extension. Happy to configure this differently if you want.

src/components/Heading/Heading.tsx
beau-charman: Again, this is to do with TypeScript and how it applies it React typings. I know there is a way to configure different, just haven't figured it out again.

beau-charman: Also, the `import * as React from 'react';` might be an artefact from when I first started the code changes, and then coincidently 'fixed it' in the rest of the time I spent on it 😅 and just didn't go back.

rollup.config.js
```diff
@@ -5,13 +5,15 @@ import copy from 'rollup-plugin-copy-assets';
...
-import postcss from 'rollup-plugin-postcss';
+import postcss from 'rollup-plugin-postcss-modules';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import { terser } from 'rollup-plugin-terser';
+import typescript from '@rollup/plugin-typescript';
 import pkg from './package.json';
 
 /* eslint-disable-next-line import/no-default-export */
 export default {
+  moduleResolution: true,
```
MkMan, line 16: Rollup doesn't recognise this option. Can I remove it?

image

package.json
MkMan: I tried moving to the latest version of 16.8 of `@types/react` but it's missing the `loading` attribute on `img` which resulted in a compilation error. Leaving this as it is for now.

beau-charman:
Going to leave the `@types/react` item for later: https://github.com/facebook/react/issues/16942

MkMan (approved):

PR #328, opened 2021-02-27
AE-171 fix tag push workflows

MkMan:
Since adding a second workflow triggered by tag pushes both tag triggered workflows have been running intermittently. I couldn't find a reason why online but I have a theory that it's having 2 workflows triggered by the same event. 

This PR combines the 2 workflows into 1. I tested it with 3 tag pushes and it ran every time, hopefully it stays that way 🤞 
image


.github/workflows/tag-push.yml
```diff
@@ -1,12 +1,49 @@
...
+
+      - name: Checkout private tools
+        uses: actions/checkout@v2
+        with:
+          repository: aesop/aesop-github-action
+          token: ${{ secrets.X_GITHUB_TOKEN }}
+          path: actions
+
+      - name: Update Gel
+        uses: ./actions/create-pr
```
kevin-ho87, line 40: How come this path uses `./actions/..` and the others use `actions/...`.
Are they just in different path locations?

MkMan: Ah 😃 
* `actions/...` refers to actions from the GH marketplace. For example, [actions/checkout](actions/checkout@v2)
* `./actions/create-pr` is referring to a [custom action](https://github.com/aesop/aesop-github-action/tree/master/create-pr) created by us (Aesop) and cloned in the folder `actions`. This is done in the previous step

kevin-ho87: 👍  makes sense

beau-charman:
How is this one going @MkMan ?

MkMan:
> How is this one going @MkMan ?

Just waiting for an approval to merge it 👍

maxlew (approved):
Sorry let this one sit for a while, looks good though 👍🏻

PR #333, opened 2021-03-02
fix: wrap document access in window check

MkMan:
Access to the `document` without a check is causing errors when Gel is used outside the browser (i.e. NextJS). This change wraps them in a check to prevent execution unless in the browser.


beau-charman (approved):

maxlew (approved):
Ah nice, I think in my hacky POC I moved them to a useEffect but it probably makes sense to just completely ignore if it runs on a server

PR #336, opened 2021-03-09
[Feature][AES-2550] Update the style of checkbox and error message

zack76:
Update the style of checkbox as well as the error messages
image
image


src/components/Checkbox/Checkbox.module.css
```diff
@@ -56,22 +56,29 @@
   &:hover {
     cursor: pointer;
   }
-
-  &.hasError {
-    border-color: var(--color-error);
```
MkMan, line 61: Was this requested by the UX team?

zack76: > Was this requested by the UX team?

Yes, the checkbox doesn't need to be in red color in this case

MkMan (approved):

src/components/Checkbox/Checkbox.module.css
```diff
@@ -56,22 +56,25 @@
...
 
 .content {
   margin-left: 10px;
   font-size: rem(14px);
   line-height: 2;
+
   &.dark {
     color: var(--color-dark-copy);
+    a {
+      color: var(--color-dark-copy);
```
kevin-ho87, line 69: you can do `color: currentColor` so it inherits the parent colour. Simpler i feel

zack76: > you can do `color: currentColor` so it inherits the parent colour. Simpler i feel

Thank you! Updated.

kevin-ho87: Just make sure it still works with the 'light/dark' theme haha

kevin-ho87 (approved):

src/components/Checkbox/Checkbox.module.css
zack76: > Just make sure it still works with the 'light/dark' theme haha

yeah, i tested it and all look good

PR #337, opened 2021-03-10
Migrate Paragraph to TS

MkMan:


src/components/Paragraph/Paragraph.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { FC } from 'react';
+import cx from 'classnames';
+import styles from './Paragraph.module.css';
+import type { ParagraphProps, ParagraphSetProps } from './Paragraph.types';
```
beau-charman, line 4: Go me thinking, should we be prefixing the `Props` with its context? I though the context would imply what the props are for, but a situation like this makes sense for the prefix.

beau-charman (approved):

src/components/Paragraph/Paragraph.tsx
MkMan: It's a good idea. I noticed when I was importing `Props` in #338 the IDE was suggesting Props from the whole repo which can be a nuisance if you pick the wrong one. Not fussed either way :) If we decide to prefix I can raise a separate PR to refactor the existing code in `develop`

beau-charman: Okay! I am sold

PR #342, opened 2021-03-10
Rrefactor to new conventions

### LINKED ISSUE #341: Prefix existing `Props` types with current context.

MkMan:
This PR refactors existing TS code to:
1. Prefix `props` with the name of the component/context they're for
2. Move css imports to the last line of imports in the file

Closes #341

### DIFF
--- a/src/components/Accordion/Accordion.types.tsx
+++ b/src/components/Accordion/Accordion.types.tsx
@@ -1,18 +1,18 @@
 import { ReactNode } from 'react';
 import type { Themes } from '~/types';
 
-type GelAccordionItem = {
+type AccordionItem = {
   content: ReactNode;
   heading: string;
   id: string;
 };
 
-type Props = {
+type AccordionProps = {
   className?: string;
   id?: string;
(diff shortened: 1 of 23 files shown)


src/components/Accordion/Accordion.types.tsx
```diff
@@ -7,12 +7,12 @@ type GelAccordionItem = {
   id: string;
 };
 
-type Props = {
+type GelAccordionProps = {
```
MkMan, line 10: I opted to add Gel to this one because we use an external library for Accordions and felt this would remove ambiguity. Happy to make it just `AccordionProps` though 🙂

beau-charman: I appreciate the concern, but maybe just `AccordionProps` in side this file the `Gel` prefix makes sense, but any other component or application consuming this component are not exposed to the external module we are using, so the prefix would be confusing most of the time.

MkMan: Done 👍

beau-charman (approved):

PR #345, opened 2021-03-15
Migrate Audio component to TS

hassanraza0054:


src/components/Audio/components/SeekButton/SeekButton.tsx
```diff
@@ -1,12 +1,18 @@
...
+import { SeekButtonProps } from './SeekButton.types';
 import styles from './SeekButton.module.css';
 
-const SeekButton = React.memo(
-  ({ copy, direction, isLoading, onClick, progressColor }) => {
+const SeekButton = React.memo<SeekButtonProps>(
+  ({
+    copy = {},
+    direction = 'forward',
+    isLoading = false,
```
hassanraza0054, line 12: Question.

When migrating to typescript, had this curious interaction. When setting the default value to the boolean **false**, it was working as intended, no error.

That said, if I set the default value to **string**, it allowed that type to go through, and extended the possible **Type** of isLoading to **string | boolean**. I assumed it would give an error for non-boolean types being assigned as default values. Maybe its a quirk with **react memo**? Not sure yet.

**String Value inserted**
image

**Boolean Value inserted**
image

beau-charman: It's *interfering* a combined type based on the 'type' and its default value. When consuming/using the function (through the component API), any other value rather than a boolean will though an error. What you are saying via authoring a `'false'` value is "only accept a boolean, however, if the value is undefined, then assign the `string` of 'false'". You IDE is just displayed that as `string | boolean`. Hopefully that makes sense.

beau-charman: > I assumed it would give an error for non-boolean types being assigned as default values

nah, you are authoring the API for this function, your type might say boolean, but you yourself extended the possible values by saying the default could be a string, but ONLY the default value.

beau-charman: https://www.typescriptlang.org/docs/handbook/type-inference.html

beau-charman: as that doc describes, typescript build a API blueprint based on explicit types and inferred values. 

Like, declaring and assigning a value doesnt throw an error in type script, as the type is inferred: 

```
const x = 1; 
```  
is the same as authoring: 
```
const x: string = 1; 
```  
the primitive value of `1` inferrs a number type

hassanraza0054: Interesting. I actually thought that once we declared the **Type** structuring, even default values would be enforced to be only boolean. But I guess that isn't the case. Thanks Beau.

src/components/Audio/Audio.tsx
```diff
@@ -1,20 +1,20 @@
...
 import { useWaveSurfer } from './Audio.customHooks';
+import type { AudioProps } from './Audio.types';
 import styles from './Audio.module.css';
 
-const Audio = ({
+const Audio: FC<AudioProps> = ({
   artistName,
   audioUrl,
   className,
   copy,
```
MkMan, line 14: I just noticed this prop is missing its default value of `{}`

src/components/Audio/Audio.tsx
```diff
@@ -1,20 +1,20 @@
...
 import cx from 'classnames';
 import { Loading } from '~/components/Loading';
 import { AudioHeader } from './components/AudioHeader';
 import { AudioFooter } from './components/AudioFooter';
 import { useWaveSurfer } from './Audio.customHooks';
+import type { AudioProps } from './Audio.types';
 import styles from './Audio.module.css';
 
-const Audio = ({
+const Audio: FC<AudioProps> = ({
```
zack76, line 10: do you have any idea why this component is use `FC` while the others are using `React.memo`

beau-charman: FC is a type

MkMan (approved):

src/components/Audio/components/PausePlayButton/PausePlayButton.tsx
```diff
@@ -1,12 +1,18 @@
...
 import { Button } from '~/components/Button';
 import { Icon } from '~/components/Icon';
+import type { PausePlayButtonProps } from './PausePlayButton.types';
 import styles from './PausePlayButton.module.css';
 
-const PausePlayButton = React.memo(
-  ({ copy, isLoading, isPlaying, onClick, progressColor }) => (
+const PausePlayButton = React.memo<PausePlayButtonProps>(
+  ({
+    copy = {},
```
beau-charman, line 10: Could we use optional chaining (`copy?.loading`) instead of declaring an empty array? Gives the impression we are going to mutate it and add methods to it

beau-charman (approved):

src/components/Audio/components/PausePlayButton/PausePlayButton.tsx
hassanraza0054: So instead of 
image

I should instead do

**No default Value**
image

**Optional Chaining**
image

Is that right?

beau-charman:
Besides the comment about the `copy` object default, this looks good to go

src/components/Audio/components/PausePlayButton/PausePlayButton.tsx
beau-charman:
```
copy={{
  downloadLabel: copy?.downloadLabel,
  downloadTitle: copy?.downloadTitle,
  loading: copy?.loading,
  pause: copy?.pause,
  play: copy?.play,
  seekBackwardTitle: copy?.seekBackward,
  seekForwardTitle: copy?.seekForward,
}}
```

beau-charman: Or, you could destructor the props and use a default object literal`{}` in context. 

```
const  { downloadLabel,
  downloadTitle,
  loading,
  pause,
  play:,
  seekBackward,
  seekForward, 
} = copy || {}; 

...

copy={{
  downloadLabel,
  downloadTitle,
  loading,
  pause,
  play:,
  seekBackwardTitle: seekBackward,
  seekForwardTitle: seekForward,
}}
```

hassanraza0054: Yup went for the first option with optional chaining! Thanks Beau. Will have a think of the second as that looks interesting too. Both are semantic choices right?

PR #346, opened 2021-03-15
Fix React Dom Server dependency in browser

### LINKED ISSUE #344: Refactor Google maps component to remove usage of server only method
The `renderToString` method used [here](https://github.com/aesop/aesop-gel/blob/develop/src/components/GoogleMap/GoogleMap.js#L205) is marked as [a server only method](https://github.com/facebook/react/blob/9198a5cec0936a21a5ba194a22fcbac03eba5d1d/packages/react-dom/src/server/ReactDOMStringRenderer.js#L12). The method uses inbuilt NodeJs module (e.g. Stream) that are not available in a browser environment. 

This hasn't been an issue thus far because Webpack on Aesop Web UI used to bundle polyfills for those Node modules. However, from Webpack v5 onwards that will no longer be the case.

Instead of shipping polyfills for those modules, it seems best to address the issue at the cause.

MkMan:
Currently, one of the dependencies of the `renderToString` method is the Node Stream module. This isn't an issue at the moment as Webpack v4 (and older) provide a polyfill for it. However, from v5 onwards we'd need to provide our own polyfill for it.

Fortunately, there is a browser friendly alternative provided in the React Dom library. Switching to it remove the `Stream` dependency from the build files as seen in the diff below. In addition to moving to the browser version, I've switched to using `renderToStaticMarkup`. This method is sufficient for our use case as we won't be hydrating the component later and thus don't need `renderToString`.

image

I deployed a Web UI branch using this Gel branch [here](https://demo-react-dom-change.acceptance.aesop.com/au). It works in IE too @kevin-ho87 
image

Closes #344

### DIFF
--- a/src/components/GoogleMap/GoogleMap.js
+++ b/src/components/GoogleMap/GoogleMap.js
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import isFunction from 'lodash/isFunction';
-import ReactDOMServer from 'react-dom/server';
+import { renderToStaticMarkup } from 'react-dom/server.browser';
 import MarkerClusterer from '@google/markerclusterer';
 import {
   HYPERLINK_STYLE_TYPES,
@@ -202,7 +202,7 @@ const GoogleMap = ({
       });
 
       const infoCard = new googleMap.maps.InfoWindow({
-        content: ReactDOMServer.renderToString(
+        content: renderToStaticMarkup(
           <InfoCard
             address={address}
             copy={{


kevin-ho87:
Does this package work with IE? 😭

MkMan:
> Does this function work with IE? 😭

Not sure 😕   I know Gel still works with IE but I don't think this component is being used at the moment so I am not able to test that it works.

zack76:
> > Does this function work with IE? 😭
> 
> Not sure 😕 I know Gel still works with IE but I don't think this component is being used at the moment so I am not able to test that it works.

kevin-ho87 (approved):

PR #349, opened 2021-03-18
[Video Component] Scroll based video tweaks

beau-charman:
This PR adds the Scroll Based Video to the Storybook and ensures a `videoRef.current.load();` on source prop changes and


MkMan (approved):
Considering the implementation changed, should one of the commits be a fix/feat to increment the version?

src/components/Video/Video.customHooks.js
```diff
@@ -7,7 +7,10 @@ const useVideoScroller = (videoRef, isScrollBasedVideo) => {
 
     if (isScrollBasedVideo && videoRefCurrent) {
       // eslint-disable-next-line no-new
-      new VideoScroller({ el: videoRefCurrent });
+      new VideoScroller({
+        el: videoRefCurrent,
+        invert: true,
```
kevin-ho87, line 12: What does `invert` here change?

beau-charman: plays the video forward instead of backwards

PR #351, opened 2021-03-22
Prepend the deploy Storybook task with build Storybook task

beau-charman:
My apologies @MkMan, in my attempts to keep the tasks names and what they do simple, I have realised that `deploy-storybook` without the prepended `build-storybook` is not a useful as I thought it would be. 

Also, would this change effect the Github Actions at all?


MkMan:
> Also, would this change effect the Github Actions at all?

We should be able to remove this [step](https://github.com/aesop/aesop-gel/blob/develop/.github/workflows/deploy-gh-pages.yml#L38-L39) then?

PR #352, opened 2021-03-22
Migrate AddToCartButton to TS

MkMan:


src/components/AddToCartButton/AddToCartButton.spec.tsx
```diff
@@ -39,11 +39,12 @@ describe('<AddToCartButton />', () => {
   const copy = fixture.copy;
 
   beforeAll(() => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
     jest.spyOn(global.console, 'error').mockImplementation(() => {});
   });
 
   afterAll(() => {
-    global.console.error.mockRestore(); // eslint-disable-line no-console
+    (global.console.error as jest.Mock).mockRestore(); // eslint-disable-line no-console
```
zack76, line 47: maybe i'm silly that i dont quite understand this. Could you please explain a bit for me? Thank you!

zack76 (approved):

src/components/AddToCartButton/AddToCartButton.spec.tsx
MkMan: Not silly 😃 

Without making this change, TS gives this error
image

The reason is that **usually** there is no `mockRestore` function on `console.error`. However, it is there now because of the jest mocking done on line 42/43 

```js
jest.spyOn(global.console, 'error').mockImplementation(() => {});
```

`as` is a Typescript construct to assert the type of something. In this case, because of the mock used above `console.error` now has jest mock functions so I used `as` to instruct TS on the type.

You can read more about it [here](https://basarat.gitbook.io/typescript/type-system/type-assertion).

PR #362, opened 2021-03-24
[AE-237] Migrate List component to TypeScript

beau-charman:
Addresses: https://aesoponline.atlassian.net/browse/AE-237


src/components/List/List.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { forwardRef } from 'react';
+import cx from 'classnames';
+import { useThemeContext } from '~/contexts';
+import type { ListProps } from './List.types';
+import styles from './List.module.css';
+
+const List = forwardRef<HTMLUListElement, ListProps>(function ListRef(
+  { className, items, listItemClassName, theme },
```
MkMan, line 8: Not sure if it matters but the theme prop used to have a default value before

MkMan (approved):

src/components/List/List.tsx
```diff
@@ -0,0 +1,26 @@
...
+  ref,
+) {
+  const currentTheme = useThemeContext(theme, 'dark');
```
beau-charman, line 11: > Not sure if it matters but the theme prop used to have a default value before

@MkMan

PR #370, opened 2021-03-25
ci: using checkout v1 and npm cache to avoid weird build issues

MkMan:
I re-ran this more than 5 times and it passed. Hoping it fixes our build issues 🤞


maxlew (approved):
I wish I remembered why I added the persists creds flag 😅 should teach me to add better commit messages.

PR #371, opened 2021-03-25
Migrate Carousel to TS

MkMan:


src/components/Carousel/Carousel.types.ts
```diff
@@ -0,0 +1,33 @@
+import type { LegacyRef } from 'react';
+import type { Themes } from '~/types';
+import type { ImageProps } from '../Image/Image.types';
```
beau-charman, line 3: candidate for the `types` directory?

MkMan: I am glad you picked up on this 🙂 I wanted to discuss it with you but forgot.
I am 50/50 on it, both locations make sense but also have drawbacks. I was also thinking we could leave it inside the Image folder but import and export it from the types directory for other components to use.

beau-charman: I misread `ImageProps` as `Image` :( alllll good to go with this one!

beau-charman (approved):

PR #373, opened 2021-03-28
[AE-241] Migrate Modal component to TypeScript

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/AE-241


src/components/Modal/Modal.types.ts
```diff
@@ -0,0 +1,14 @@
+import type { Themes } from '~/types';
+
+type ModalProps = {
+  children: React.ReactNode;
```
MkMan: We can leave this one out, the FC type declares it

src/components/Modal/Modal.types.ts
```diff
@@ -0,0 +1,14 @@
...
+  className?: string;
+  copy?: {
+    close?: string;
+  };
+  isVisible: boolean;
+  onClose: () => void;
```
MkMan, line 9: These are the same between `Modal` and `ModalBody`, should we share their types instead of duplicate its definition in both places? `copy` specifically is only used to be passed into `ModalBody` as a prop.

src/components/Modal/Modal.types.ts
beau-charman: Yeah I pushed a commit just after the PR

src/components/Modal/Modal.types.ts
beau-charman: Yeah happy to do that

MkMan (approved):

beau-charman:
What are your thoughts on the prop types @MkMan. I wanted to keep them seperate and extendable, but yeah needed to explicitly document where and why some props are shared

MkMan:
> What are your thoughts on the prop types @MkMan. I wanted to keep them seperate and extendable, but yeah needed to explicitly document where and why some props are shared

That's a really tough one... I am not sure 😕 There are many ways to achieve what we want and the good thing is there is no right or wrong answer. I think it just comes down to what's easiest to build on in the future and what makes the most sense. 

I see what you did in the latest commit moving to `ModalAndModalBodyProps`.

Another approach we could try is seeing that 

```TS
type ModalProps = ModalAndModalBodyProps & ModalBodyCopy;

// and

type ModalBodyProps = ModalBodyCopy & ModalAndModalBodyProps;

// are the same
```

we can 
```TS
// src/components/Modal/components/ModalBody/ModalBody.types.ts
type ModalBodyProps = {
  className?: string;
  copy?: {
    close?: string;
  };
  isVisible: boolean;
  onClose: () => void;
  theme?: Themes;
};

export type { ModalBodyProps };
```
```TS
// src/components/Modal/Modal.types.ts
import type { ModalBodyProps } from './components/ModalBody/ModalBody.types';

type ModalProps = ModalBodyProps;

export type { ModalProps };
```

Honestly, I am not sure and also not fussed. We can go with whatever works now and if a need arises in the future to change it we can do that 🙂

beau-charman:
`type ModalProps = ModalBodyProps;` throws an assignment error ("`ModalBodyProps` is a type but is being used as a value") 

So I was exporting just as `export type { ModalBodyProps as ModalProps };`, but yeah, wanted to be able to extend the types. 

I will leave as is

PR #379, opened 2021-03-30
fix: refactor css module extends

beau-charman:
seems our implementation of rollup is not parsing class base extends, only silent extends. This PR addresses that. 

We have no need for a `.paragraph` class anyway, just need the classes style values.


hassanraza0054 (approved):

MkMan (approved):

hassanraza0054:
> seems our implementation of rollup is not parsing class base extends, only silent extends. This PR addresses that.
> 
> We have no need for a `.paragraph` class anyway, just need the classes style values.

Interesting to note the way this was found out.

https://aesoponline.atlassian.net/browse/CON-136

Working on the above ticket, I found that we had a general UI requirement for paragraphs to have line breaks in the form of spaces between paragraphs. That said, the links shared below show that, while we had a programmatic way to solve this in the form of **extending** paragraph functionality on the paragraph base class, the extends wasn't going through for styles.

Instead, even the working example below is a bit misleading as the content had line spaces hard coded. So in actuality, our programmatic line breaks weren't going through till this PR change. Thanks a ton @beau-charman for help on this 🙏 

Misleading Paragraph line spacing [Content hard coded]
https://www.aesop.com/au/r/sublime-eudaimonia/

image

image

Actual Paragraph Behaviour [As at 30th March 2021]

image

image

PR #380, opened 2021-03-30
chore: disable update_web_ui workflow

MkMan:
As the `develop` branch is no longer being used and will be deleted soon I thought we should disable this part of the workflow until we figure out how to update Gel on Web UI.

### DIFF
--- a/.github/workflows/tag-push.yml
+++ b/.github/workflows/tag-push.yml
@@ -3,7 +3,6 @@ on:
   push:
     tags:
       - v*
-  workflow_dispatch:
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-push.yml
```diff
@@ -10,6 +10,7 @@ env:
 
 jobs:
   update_web_ui:
+    if: ${{ false }}
```
MkMan, line 13: Does this look ok, @maxlew ?

beau-charman (approved):

maxlew (approved):

PR #381, opened 2021-03-30
Reconfigure eslintrc to run typescript eslint as priority

beau-charman:
+ refactor scripts to remove new linting warnings and errors


MkMan (commented):
Is there a concern these changes will change how the built code works? If not then we probably don't need to increment the version through a `feat` commit 🙂

beau-charman:
> Is there a concern these changes will change how the built code works? If not then we probably don't need to increment the version through a `feat` commit 🙂

Screen Shot 2021-04-01 at 8 20 02 am

I will change the commit message

PR #388, opened 2021-04-06
feat(ae-219): migrate DefinitionListWithProductExtract to TS

MkMan:


src/components/DefinitionListWithProductExtract/DefinitionListWithProductExtract.type.ts
```diff
@@ -0,0 +1,19 @@
...
+  productBottomBorder?: boolean,
+  productItemNumber?: number,
+  product?:{ // TODO: use definition in ProductExtract after it's migrated
+    image?: any,
+    name?: string,
+    url?: string,
+  },
+  theme?: Themes,
+  works?: string,
+} & Pick<DefinitionListProps, "items">;
```
zack76, line 17: A question: how does this work?

MkMan: It will create an `items` type that is the same as its type defined in `DefinitionListProps`. I did this because the `items` prop gets passed directly into `DefinitionList` therefore they must be the same type. Instead of redefining it or exporting it from `DefinitionList` and importing and using it here I preferred to use this to avoid unnecessary duplication or exports.

I put together a simplified example [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAksEFsoF4oG8BQVtQJYBMAuKAZ2ACdcA7Acyx3whIGNizLaMBfDDUSKAGEA9gjABGAArlhYEinT1szADYBDEiQByahBDYVqdHHngISxOIgDaAXW69+0EWIBM02fNSYT+NcDUAFSZgACUIADMDDjouKAAyKElcZgBrAB4XCQ85ABooAHJcMxICgD4Abl5mYSoyKBq3HIshUTB3GTkFHwZ-IJDwqMK1fHxyJhIAWgjcCBV8AtylU0QW6zQ8IkKSUQhJgkWoRhZiAsCAC2gZufwoYGEoCCp4cigRsYmmAqguey4KoA) and you can find the official documentation [here](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys).

zack76 (approved):

PR #391, opened 2021-04-08
Refactor Grid Overlay from Storybook to TypeScript and expose for other libraries

beau-charman:


beau-charman:
@MkMan , would you have any suggestion on being able to expose `GridOverlay` to consuming libraries, but not in production builds? Thank you for any thoughts here

MkMan:
> @MkMan , would you have any suggestion on being able to expose `GridOverlay` to consuming libraries, but not in production builds? Thank you for any thoughts here

I have a couple of ideas, neither of which I tried but here goes 🙂 

### 1. New build output ONLY for dev components
It will look something like:
1. New `index.ts` file export components that are to be used only during development
2. Updating our rollup config to generate build files for this index file separately from the other prod-ready components
3. Updating package.json (if needed) to include the generated files 
4. In Web UI, to use these dev only components we'd need to add an import like `import { Grid } from 'aesop-gel/dev';`


### 2. New build output but Gel decides which to provide based on env
It will look something like:
1. Rename existing built files to indicate they're for production env. e.g. `index.prod.js`
2. New `index.ts` file export prod **and** dev components 
3. Updating our rollup config to generate build files for this index file separately from the other prod-ready components
4. Add a new `index.js` file that will decide which file (from `index.prod.js` and `index.dev.js`) to return when Gel is imported based on an env variable (probably NODE_ENV)
5. In Web UI we import Gel the same way for both production and development component and if our set up is correct we'll get the development components locally but they won't be there in non development environments.

### Recommendation
Personally, I prefer the first option for the following reasons:
1. It's easier to implement
2. We're less likely to miss/forget an import that looks like `import { Grid } from 'aesop-gel/dev';` than `import { Grid } from 'aesop-gel';`
3. No reliance on Node environment variables

@maxlew might also have some ideas 🙂

PR #393, opened 2021-04-08
[TypeScript Migration][AE-239] Loading Component

beau-charman:
JIRA: https://aesoponline.atlassian.net/browse/AE-239


src/components/Loading/Loading.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { forwardRef } from 'react';
+import cx from 'classnames';
+import { useThemeContext } from '~/contexts';
+import type { LoadingProps } from './Loading.types';
+import styles from './Loading.module.css';
+
+const Loading = forwardRef<HTMLDivElement, LoadingProps>(function LoadingRef({
```
MkMan, line 7: Should the ref type be `HTMLSpanElement` instead?

src/components/Loading/Loading.tsx
```diff
@@ -0,0 +1,38 @@
...
+  className,
+  isLoading,
+  shouldFillSpace = false,
+  size = 'medium',
+  theme,
+}) {
```
MkMan, line 13: Don't forget to capture and use the `ref` argument 🙂

src/components/Loading/Loading.tsx
beau-charman: Thank you for the pick up. I might just remove the `forwardRef`.

src/components/Loading/Loading.tsx
beau-charman: yeah, as above I should havent even have included it

MkMan (approved):

PR #398, opened 2021-04-14
[Feature][AES-116] Refactor JavaScript and CSS breakpoints and viewports and add new image breakpoints

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/CON-116, 

The way breakpoints and viewports (media queries) are declared and used needed to be re-thought and refactored to be scaleable and maintainable.


src/utils/viewports/viewports.ts
```diff
@@ -0,0 +1,16 @@
+import { isInBrowser } from '~/utils/environment';
+import { VIEWPORTS } from '~/constants';
+import type { ViewportNames } from '~/types';
+
+const isViewport = (viewport: ViewportNames): boolean => {
+  /** @TODO remove this once TypeScript migration is complete */
+  if (!VIEWPORTS.get(viewport)) {
+    throw new Error(`The viewport '${viewport}' does not exist`);
+  }
```
MkMan, line 9: I think we can still keep this after the migration 🙂  If the consuming app isn't using TS it would be good to know something went wrong.

On a somewhat related note, Gel isn't generating type definitions. I think even if the consuming app is using TS they won't know what value to provide for the `viewport` argument

src/constants/breakpoints.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const BREAKPOINTS: Map<BreakpointNames, BreakpointValues> = new Map([
+  ['xs', { minWidth: 0, maxWidth: 320 }],
+  ['sm', { minWidth: 321, maxWidth: 639 }],
+  ['md', { minWidth: 640, maxWidth: 1024 }],
+  ['lg', { minWidth: 1025, maxWidth: 1219 }],
+  ['xl', { minWidth: 1220, maxWidth: 1599 }],
+  ['x2l', { minWidth: 1600, maxWidth: 1919 }],
+  ['x3l', { minWidth: 1920, maxWidth: 2399 }],
+  ['x4l', { minWidth: 2400 }],
```
MkMan, line 16: That's a lot of breakpoints 😄

MkMan (approved):
Thanks for converting those files to TS 🙂

src/constants/breakpoints.ts
beau-charman: I know :(

zack76 (approved):

beau-charman:
Screen Shot 2021-04-19 at 9 28 57 am

at least I added less then I took away @MkMan

PR #404, opened 2021-04-20
Typescript: add shared component types

MkMan:
Moving away from `React.FC` (refer to discussion [here](https://github.com/aesop/aesop-gel/pull/397#discussion_r614417432)) we thought it would be handy to create our types instead of manually needing to specify the return type for components (which would almost always be the same).

I created 2 new types: `ComponentWithChildren` and `ComponentWithoutChildren` and migrated Transition to use `ComponentWithChildren` to demo it.

Any feedback is appreciated 😃

### DIFF
--- a/src/types/Component.types.ts
+++ b/src/types/Component.types.ts
@@ -1,0 +1,15 @@
+/* eslint-disable @typescript-eslint/ban-types */
+import type { ReactElement, ReactNode } from 'react';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type DefaultReactComponentReturn = ReactElement<any, any> | null;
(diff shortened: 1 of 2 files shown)


beau-charman:
This is Cool. Transition is probs a bad example for the rule, but a good one to illustrate the exception (as it's `children` props is of the type `JSX.Element`)

MkMan:
> This is Cool. Transition is probs a bad example for the rule, but a good one to illustrate the exception (as it's `children` props is of the type `JSX.Element`)

Ah yeah, good point. I'll drop the Transition commit

zack76 (approved):

beau-charman (approved):
Can we pull this in? :)

PR #405, opened 2021-04-20
Migrate DoubleMedia to TS

MkMan:


src/components/DoubleMedia/DoubleMedia.spec.tsx
```diff
@@ -20,7 +22,8 @@ describe('<DoubleMedia />', () => {
         mediaTwo={{
           heading: DoubleMediaFixture.title,
           link: DoubleMediaFixture.link,
-          type: DoubleMediaFixture.imageTwo.type,
+          type: DoubleMediaFixture.imageTwo
+            .type as DoubleMediaProps['mediaTwo']['type'],
```
kevin-ho87, line 26: how does this `as` work?

MkMan: It's [type assertion](https://basarat.gitbook.io/typescript/type-system/type-assertion). I am overriding the type that TS inferred for that property (in this case a `string`) and saying it's actually the same type as `DoubleMediaProps['mediaTwo']['type']`.

Without it, the compiler gives an error 
image

kevin-ho87: ahh icic

kevin-ho87 (approved):

zack76 (approved):

PR #409, opened 2021-04-21
Migrate useHasMounted to TS

MkMan:


zack76 (approved):

src/customHooks/useHasMounted/useHasMounted.ts
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react';
 
-const useHasMounted = () => {
+const useHasMounted = (): boolean => {
```
zack76, line 3: a quick question: why you don't define return type here?

PR #418, opened 2021-04-27
Expose Typescript declarations in build directory

beau-charman:
- Use `tsc` to generate declaration files
- Rollup all type declarations with `rollup-plugin-dts`
- Ensure all paths that reference index files are to `index.ts` files 

TODO
- src/components/DefinitionListWithProductExtract/DefinitionListWithProductExtract.tsx remove ts ignores
~~Generate css type files with rollup-plugin-postcss-modules and move them to `styles.d.ts` https://github.com/Quramy/typed-css-modules~~


tsconfig.json
```diff
@@ -9,5 +11,7 @@
     "esModuleInterop": true,
     "jsx": "react",
     "plugins": [{ "name": "typescript-plugin-css-modules" }]
-  }
+  },
+  "include": ["src"],
+  "exclude": ["**/*.spec.*", "**/*.stories.*", "**/*.md*"]
```
MkMan, line 16: Should we move the excludes array specified in rollup config here as well? That way all the tsconfig is in 1 place

src/components/DefinitionListWithProductExtract/DefinitionListWithProductExtract.tsx
```diff
@@ -29,16 +30,24 @@ const DefinitionListWithProductExtract: FC<DefinitionListWithProductExtractProps
...
+        // @ts-ignore
         isVisible={isVisible}
+        // @ts-ignore
         itemNum={productItemNumber}
+        // @ts-ignore
         product={product}
+        // @ts-ignore
         theme={theme}
+        // @ts-ignore
         works={works}
```
MkMan, line 46: 😱

MkMan (commented):
Really nice 👌 
One thing I recommend, I think it's best if the changes to styles are excluded from this PR. They're not related (as far as I can tell) and there should be no behaviour change as a result of this PR.

tsconfig.json
beau-charman: This is a config specifically for generating TS declaration files

src/components/DefinitionListWithProductExtract/DefinitionListWithProductExtract.tsx
beau-charman: Yep. Happy to explain over a call, but basically because we still have js files that need to be typed for that the root level index.ts files work - however the ProductExtract component is one that is wrapped in a Ref, and that IntrinsicType + ReactRef type doesnt have these other props on it. only this component. only temporary.

beau-charman:
> Really nice 👌
> One thing I recommend, I think it's best if the changes to styles are excluded from this PR. They're not related (as far as I can tell) and there should be no behaviour change as a result of this PR.

after trying to implement https://github.com/Quramy/typed-css-modules, the tool was correctly throwing an error that these class names did not exist. they are the type of silent however that usually throws errors in IE - so I'd prefer to get rid of it now.

src/components/DefinitionListWithProductExtract/DefinitionListWithProductExtract.tsx
MkMan: Fair enough. If possible, can you please add a comment/TODO so we remove them in the future once they're migrated?

MkMan (approved):

src/components/DefinitionListWithProductExtract/DefinitionListWithProductExtract.tsx
beau-charman: I can do that

PR #425, opened 2021-05-02
[Typescript Configuration] Create dedicated config for generating type definitions

beau-charman:
+ Addresses "react can only be default-imported using the 'esModuleInterop' flag" issue
+ Add transform plugin to ensure alias paths are correctly transformed

for example: 

```
import { Themes as Themes$1 } from '~/types';
```

is replaced by

```
declare type Themes = 'dark' | 'light';
```


package.json
```diff
@@ -27,7 +27,7 @@
     "test": "jest",
     "test-coverage": "open coverage/lcov-report/index.html",
     "test-watch": "jest --watch --verbose",
-    "type-definitions": "tsc --declaration --emitDeclarationOnly --allowJs"
+    "type-definitions": "ttsc --project tsconfig.definitions.json --declaration --emitDeclarationOnly --allowJs"
```
MkMan, line 30: Could this script just be `ttsc --project tsconfig.definitions.json` instead? The tsconfig file already has `"declaration": true,` and the other 2 arguments can be moved to the config file

MkMan (approved):

package.json
beau-charman: yeah still refactoring

PR #429, opened 2021-05-07
Remove scope case rule

maxlew:
I know the commit says adding 'scope-case' rule, but I'm adding it to remove it...

Basically adding an exception for commitlint so that the `scope` part of the commit message can be whatever case we want. This makes it a bit easier for using it to tie commits to Issue numbers. It also brings it in line with other repositories that we're using commitlint with. 

```
feat(ABC-123): adding neat feature
```


MkMan (approved):

beau-charman (approved):
🙌

PR #431, opened 2021-05-08
[Feature][AE-228] Migrate `Hidden` to Typescript

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/AE-228


src/components/Hidden/Hidden.types.ts
```diff
@@ -0,0 +1,9 @@
+type HiddenProps = {
+  children: React.ReactNode;
+  isLarge: boolean;
+  isMedium: boolean;
+  isSmall: boolean;
+  isXLarge: boolean;
```
MkMan, line 6: Are all of these required? Originally, they were optional.

beau-charman: they are optional. thank you for the pick up

PR #433, opened 2021-05-09
[AE-265] Migrate `TwoColumnList` to Typescript

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/AE-265


src/components/TwoColumnList/TwoColumnList.tsx
```diff
@@ -0,0 +1,84 @@
...
+import cx from 'classnames';
+import { useThemeContext } from '~/contexts';
+import { Heading } from '~/components/Heading';
+import { Hyperlink } from '~/components/Hyperlink';
+import { List } from '~/components/List';
+import type { ListItem } from '~/components/List/List.types';
+import type { TwoColumnListProps, ColumnProps } from './TwoColumnList.types';
+import styles from './TwoColumnList.module.css';
+
+const TwoColumnList = forwardRef<any, TwoColumnListProps>(
```
MkMan, line 11: The ref is assigned to a div, would `HTMLDivElement` be a suitable type to use instead of any?
```suggestion
const TwoColumnList = forwardRef<HTMLDivElement, TwoColumnListProps>(
```

beau-charman: Very true. context switching from the other PR that was similar must have confused me. thank you

src/components/List/List.spec.tsx
```diff
@@ -1,16 +1,8 @@
-import React from 'react';
-import { render } from '@testing-library/react';
-import { List } from './List';
-import { ListFixture } from './List.fixture';
+// import React from 'react';
+// import { render } from '@testing-library/react';
+// import { List } from './List';
+// import { ListFixture } from './List.fixture';
```
zack76, line 4: if these need to be deleted as well? I guess they will still be needed once `it.todo` is finished

zack76 (approved):

PR #436, opened 2021-05-11
[AE-247] Migrate `ProductDetailHeader` to Typescript

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/AE-247

+ Remove imperative prop values
+ Remove heading and transition const in favour for types
+ Rebuild type declarations
+ Migrate Product Excerpt as ref type errors were exhausting


src/constants/transitions.ts
```diff
@@ -1,9 +0,0 @@
-export const TYPE = {
-  FADE: 'fade',
-  SHIFT_IN_DOWN: 'shiftInDown',
-  SHIFT_IN_LEFT: 'shiftInLeft',
-  SLIDE_DOWN: 'slideDown',
-  SLIDE_RIGHT: 'slideRight',
-  SLOW_FADE: 'slowFade',
-  ZOOM: 'zoom',
-};
```
zack76, line 9: a bit curious why we stop using it?

zack76 (approved):

src/constants/transitions.ts
beau-charman: the animation type defined in the Transition component is explicit / self documenting enough - having this constant defined for what will be just a prop type is superfluous

PR #438, opened 2021-05-12
[Feature/PUR-239]: variant querystring pageload preselect

kevin-ho87:
**Ticket**
https://aesoponline.atlassian.net/browse/PUR-239

**Description**
- Query string `?variant=` to preselect variant in variant hook
- Fallback to first variant in array if no query string found


src/contexts/VariantSelectContext/VariantSelectStore.ts
```diff
@@ -2,11 +2,28 @@ import { useEffect, useState } from 'react';
...
 import type { Variant } from '~/types';
 
+function getQueryString(name) {
+  const result =
+    location && location.href.match(new RegExp(`[?&]${name}=([^&]+)`, 'i'));
+
+  if (result == null || result.length < 1) return '';
+
+  return decodeURIComponent(result[1]);
+}
```
MkMan, line 12: I am conflicted on this, not sure if it's better to write our own query string parsing code or use an external library like [query-string](https://www.npmjs.com/package/query-string). Thoughts?

If you'd prefer to go with an in-house implementation, can you please move this out to be a utility function and add unit tests?

src/contexts/VariantSelectContext/VariantSelectStore.ts
```diff
@@ -2,11 +2,28 @@ import { useEffect, useState } from 'react';
...
+
 const useVariantSelectStore = (variants = []) => {
   const [selectedVariant, setSelectedVariant] = useState({});
 
   useEffect(() => {
-    setSelectedVariant(variants[0]);
+    const queryStringVariant = getQueryString('variant');
+    const getQueryStringVariant = variants.find(
+      variant => variant.sku === queryStringVariant,
+    );
```
MkMan, line 21: The name `getQueryStringVariant` is a bit misleading, it sounds like a function you call to get the query string variant. I am not great at naming but would `variantFromQueryString` match what this variable holds?

src/contexts/VariantSelectContext/VariantSelectStore.ts
```diff
@@ -2,11 +2,28 @@ import { useEffect, useState } from 'react';
...
+    if (getQueryStringVariant) {
+      setSelectedVariant(getQueryStringVariant);
+    } else {
+      setSelectedVariant(variants[0]);
+    }
```
MkMan, line 26: Would this achieve what you're after?
```suggestion
    setSelectedVariant(getQueryStringVariant ?? variants[0);
```

kevin-ho87: ah yes thanks! i completely forgot about `??` operator

src/contexts/VariantSelectContext/VariantSelectStore.ts
kevin-ho87: yep that makes more sense

src/contexts/VariantSelectContext/VariantSelectStore.ts
kevin-ho87: updated

src/contexts/VariantSelectContext/VariantSelectStore.ts
kevin-ho87: updated

src/contexts/VariantSelectContext/VariantSelectStore.ts
kevin-ho87: done, installed the package

package.json
```diff
@@ -63,6 +63,7 @@
     "lodash": "^4.17.11",
     "moment": "^2.24.0",
     "prop-types": "^15.7.2",
+    "query-string": "^5.1.1",
```
MkMan, line 66: Please add it to the peer dependencies as well 🙏

kevin-ho87: done

MkMan (approved):

PR #443, opened 2021-05-24
Offer GEL as separate bundles

MkMan:
# Problem statement

Consider a new front end codebase that wishes to use the Button component in GEL.

In addition to importing the Button in the JS/TSX code, GEL's styles need to be imported in the consuming app's main css file.
```css
@import 'aesop-gel/dist/styles.css';
```

The current bundle structure and consumption steps causes the following issues:
1. Importing all of the css generated by GEL just to use a Button. This adds approximately 100kb to the app's bundle size.
2. Gel is not tree shakable which results in importing all of GEL's JS just to use a Button. This adds approximately 400kb to the app's bundle size.
3. Needing to manually import css. To elaborate, this is not required as the css can be left in JS to be injected in the browser.

The following is the bundle analysis of a Hello World app built using Create React App.
image

# Proposal and PR walkthrough

This PR adds new Rollup configurations to bundle each component/hook/util/etc separately **leaving** their css in the generated file. This eliminates the need to manually import GEL's css and makes component consumption as follow
```js
// import { Button } from 'aesop-gel'; // old
import { Button } from 'aesop-gel/dist/components/Button'; // new
```

This change is **not** a breaking change; the current GEL bundle structure will be retained for backward compatibility with Aesop-Web-UI.

The new structure was tested and worked in the following applications:
* CRA (with TS)
* NextJS (with TS)
* Web UI

The core changes in this PR (changes outside `/dist`) can be easily accessed [here](https://github.com/aesop/aesop-gel/pull/443/commits/e903ca0bd7ded875fb8580a3d991f2dd894f4580).

## Benefits

The following is the bundle analysis in the **same** Hello World app used before
image

1. A whopping reduction of **600kb** for the same functionality
2. Less integration steps (css)
3. Not including non-SSR friendly code, something @maxlew has faced with a couple of GEL components. For example, the GoogleMaps component isn't SSR compatible, that won't matter unless it's explicitly imported. Importing a Button won't cause NextJS to have compile errors from components not in use.


maxlew (approved):
Looks good! Are the main changes the ones in the rollup config files? Wondering if we should be avoiding commiting the `dist` folder to make these PRs a bit easier.

MkMan:
> Are the main changes the ones in the rollup config files?
Yeah, you can find them in this [commit](https://github.com/aesop/aesop-gel/pull/443/commits/e903ca0bd7ded875fb8580a3d991f2dd894f4580) the rest are generate files.

> Wondering if we should be avoiding commiting the dist folder to make these PRs a bit easier.
I am all for excluding generated code from version control 🙂

beau-charman (approved):

andrwchang:
I was going to ask the same. we can exclude the dist from being committed. Otherwise this is awesome!

andrwchang (approved):

PR #445, opened 2021-05-25
[CON-386] Migrate `SectionHeading` to TypeScript

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/CON-386


src/components/SectionHeading/SectionHeading.spec.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
MkMan, line 1: Nice! Thanks for improving the tests 👌

MkMan (approved):

PR #447, opened 2021-05-26
[CON-380] Migrate `SecondaryMessage` to TypeScript

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/CON-380


src/components/SecondaryMessage/SecondaryMessage.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+            link: {
+              title: 'SecondaryMessage link title',
+              url: 'SecondaryMessage link url',
+            },
+          },
+        ]}
+      />,
+    );
+
+    screen.getByRole('link', { name: 'SecondaryMessage link title' });
```
MkMan, line 42: I share Kent's [opinion](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-get-variants-as-assertions) on this but like him I am also not fussed. Up to you 🙂

src/components/SecondaryMessage/SecondaryMessage.spec.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { SecondaryMessage } from './SecondaryMessage';
+import { SecondaryMessageFixture } from './SecondaryMessage.fixture';
+
+describe('<SecondaryMessage />', () => {
```
MkMan, line 6: Thanks for improving the tests 🙂

src/components/SecondaryMessage/components/Message/Message.spec.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { render, screen } from '@testing-library/react';
+import { Message } from './Message';
+import { MessageFixture } from './Message.fixture';
+
+describe('<Message />', () => {
+  it('should present UI with heading copy', () => {
+    render(<Message copy="message copy" heading="heading copy" />);
+
+    expect(
+      screen.getByRole('heading', { name: 'heading copy' }),
```
MkMan, line 11: I wanted your thoughts on this. Would it be better to mock the Heading and Link and assert that the mocks are called with specific props? Currently, this is testing the implementation of the Heading component (that it renders a `<h>`).

src/components/SecondaryMessage/SecondaryMessage.spec.tsx
beau-charman: ah yeah, so do I - thank you for the catch

src/components/SecondaryMessage/components/Message/Message.spec.tsx
beau-charman: I see what you mean, but I just want to assert that it 'should present UI with heading copy', hence explicitly checking that the resulting ui contains `'heading', { name: 'heading copy' }`

MkMan (approved):

PR #450, opened 2021-05-27
[CON-385] Migrate `ProductCommerce` to Typescript

beau-charman:
Addresses https://aesoponline.atlassian.net/browse/CON-385

Requires PR into Web UI to update dispatch types: https://github.com/aesop/aesop-web-ui/pull/3829


src/contexts/AddToCartContext/AddToCartContext.types.ts
```diff
@@ -1,12 +1,34 @@
...
+type AddToCartActionType = 'fail' | 'fetching' | 'success';
+
+type AddToCartAction = {
+  type: AddToCartActionType;
+  payload?: string;
+};
+
+type OnAddToCartClick = (
+  sku: string,
+  addToCartDispatch: React.Dispatch<AddToCartAction>,
```
MkMan, line 10: How does this work without importing React 🤔 ?

MkMan (approved):

src/contexts/AddToCartContext/AddToCartContext.types.ts
beau-charman: Great catch.

I am guessing because of `"jsx": "react",` Typescript assumes React as a global, but I will have to look that up. 

I will add the import to be explicit. I will also need to make a PR to `aesop-code-styles` as I have lost a lot of lint warnings for console log statements, unused variables and non-imported packages.

MkMan: That's strange 🤔 The absence of a warning on missing React import might be because of React 17. But the rest work fine on my IDE. Have you tried restarting your IDE? Maybe the eslint service needs a restart to work with the new rules.

image

beau-charman: ah okay, thank you for the heads up

PR #453, opened 2021-05-30
[CON-387-402] Migrate `SubNav` and `Select` to Typescript

beau-charman:
Addresses: 
- https://aesoponline.atlassian.net/browse/CON-387
- https://aesoponline.atlassian.net/browse/CON-402


src/components/Select/Select.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+            label: 'item label 1',
+            value: 'item value 1',
+          },
+          {
+            id: 'item-id-2',
+            label: 'item label 2',
+            value: 'item value 2',
+          },
+        ]}
+      />,
```
MkMan, line 54: What do you think about moving this out to a component and reusing it in both tests to eliminate duplication?

src/components/Select/Select.stories.mdx
```diff
@@ -1,38 +1,72 @@
 import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs/blocks';
 import * as knobs from '@storybook/addon-knobs';
+import { StorybookGridOverlay } from '~.storybook/components';
 import { State } from '~.storybook/State';
 import { Select } from './Select';
 import { SelectFixture } from './Select.fixture';
 
-<Meta title="Form Elements/Select" component={Select} />
+<Meta title="Components/Select" component={Select} />
```
MkMan, line 8: This was to group components that are form elements (inputs) together in Storybook so they're easier to see.

zack76 (approved):

src/components/Select/Select.spec.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { Select } from './Select';
+
+describe('<Select />', () => {
```
MkMan, line 6: A couple of tests have been left out. Can you please include them or write similar tests?

src/components/SubNav/SubNav.spec.tsx
```diff
@@ -0,0 +1,185 @@
...
+            children: 'link children 1',
+            url: 'link url 1',
+          },
+          {
+            id: 'link id 2',
+            style: 'External No Icon Link',
+            children: 'link children 2',
+            url: 'link url 2',
+          },
+        ]}
```
MkMan, line 26: Is it possible to move the links out so they can be reused? Could be just 2 objects, 1 for each link and then each component uses an array of 1 or more of them depending on the test.

src/components/TwoColumnLayout/TwoColumnLayout.stories.mdx
```diff
@@ -8,6 +8,7 @@ import { Image } from '~/components/Image';
 import { SubNav } from '~/components/SubNav';
 import { SubNavFixture } from '~/components/SubNav/SubNav.fixture';
 import { TwoColumnList } from '~/components/TwoColumnList';
+import { TwoColumnListFixture } from  '~/components/TwoColumnList/TwoColumnList.fixture';
```
MkMan, line 11: Is this used?

src/components/Select/Select.stories.mdx
beau-charman: I actually didn't change this :/ not sure why is was changed in the commit :(

src/components/TwoColumnLayout/TwoColumnLayout.stories.mdx
beau-charman: Don't think so - annoying that the commit linter stopped picking these up 😞

MkMan: I don't think we have linting rules for `.mdx` files.

beau-charman: you are very right, just that this is happening across all files for me at the moment

beau-charman: and yes it is, line 64

```
<TwoColumnList
  leftColumn={TwoColumnListFixture.leftColumn}
  rightColumn={TwoColumnListFixture.rightColumn}
  theme={theme}
/>
```

src/components/Select/Select.spec.tsx
beau-charman: Oh my gosh, https://github.com/aesop/aesop-gel/blob/develop/src/components/Select/Select.spec.js thank you for the pick up

src/components/Select/Select.spec.tsx
beau-charman: yeah, happy to abstract the `options` out of the test. I like the idea of being explicit and seeing the set up and assertion in the same `it()` block, but agree that the duplication is a bit much.

src/components/TwoColumnLayout/TwoColumnLayout.stories.mdx
MkMan: Hmmmm 🤔  How/why was it working before when it wasn't imported 😂 
Good pickup 👍

src/components/SubNav/SubNav.spec.tsx
beau-charman: same idea as the `Select` component: "yeah, happy to abstract the `options` out of the test. I like the idea of being explicit and seeing the set up and assertion in the same `it()` block, but agree that the duplication is a bit much"

src/components/TwoColumnLayout/TwoColumnLayout.stories.mdx
beau-charman: tenor

src/components/SubNav/SubNav.spec.tsx
MkMan: Fair enough, I see what you mean 🙂

MkMan (approved):

src/components/Select/Select.stories.mdx
beau-charman: ahhhhhhh cause I used `generate-component` 😭

PR #460, opened 2021-06-03
Consideration to Develop (week ending June 4th)

beau-charman:
`HorizontalProductDisplayAccordion` Typescript mMigration
https://aesoponline.atlassian.net/browse/CON-414

Candles 'expander' module cropping text out of view fix
https://aesoponline.atlassian.net/browse/CON-157

Review URL: https://consideration.aesop.com/us/r/othertopias-test/?contentful=true&contentful_env=playground&preview=true


MkMan (approved):
Approving this so I don't hold up the ACC drop. Trusting you that components still work 😃

PR #461, opened 2021-06-03
[Feature/pur-158-pdp-klarna]: Klarna Widget on Product Page

kevin-ho87:
**Ticket**
https://aesoponline.atlassian.net/browse/PUR-158

**Desccription**
Klarna Widget on Product Page (ProductDetailHeader)
- Allow slot below ATC to have child component. prop = `paymentWidget`
- logic to add border (styling) depending on layout and sibling components

image


zack76:
I'm just curious if the script-src will change "https://eu-library.klarnaservices.com/lib.js". I believe this one is only for EU regions

kevin-ho87:
> I'm just curious if the script-src will change "https://eu-library.klarnaservices.com/lib.js". I believe this one is only for EU regions

That is handled in Web-UI. "https://eu-library.klarnaservices.com/lib.js" is the only library url available at this stage. If it updates in future, I can update it if needs be.

zack76 (approved):

MkMan (approved):

PR #464, opened 2021-06-10
[CON-426] Add `IEErrorContext`

beau-charman:
To address https://aesoponline.atlassian.net/browse/CON-426, the PR creates an IE Error Context that can consume an error provided by a parent provider, specifically a `IndexSizeError` to gracefully handle IE video resolution errors. 

Dependency of: https://github.com/aesop/aesop-web-ui/pull/3848


src/components/Video/Video.tsx
```diff
@@ -81,27 +84,20 @@ const Video = forwardRef<HTMLDivElement, VideoProps>(function VideoRef(
 
   useEscapeKeyListener(stopVideo);
 
-  const captionsTrack = videoRef.current?.textTracks[0];
-
-  React.useEffect(() => {
+  useDeepCompareEffect(() => {
```
kevin-ho87, line 87: is this deep compare useEffect called again only when `sizes` change?

beau-charman: `useEffect` only compares primitives, as `sizes` is always a new reference to an object on update (and useProgress is causing constant renrenders), the old effect would run constantly, even is `small`, `medium` etc members of that object stay the same

kevin-ho87: got it

kevin-ho87 (approved):

src/components/Video/Video.stories.mdx
```diff
@@ -6,6 +6,8 @@ import { MediaWithContentFixture } from '~/components/MediaWithContent/MediaWith
 import { MediaWithContent } from '~/components/MediaWithContent';
 import { Video } from './Video';
 
+import { IEErrorContextProvider } from '~/contexts/IEErrorContext';
```
MkMan, line 9: Is this import being used in the file?

src/contexts/IEErrorContext/IEErrorContext.types.ts
```diff
@@ -0,0 +1,8 @@
+import type { ReactNode } from 'react';
+
+type IEErrorContextProps = {
+  children: ReactNode;
+  error?: string | undefined;
```
MkMan, line 5: Is the ` | undefined` needed? I think the fact that `error` is optional implies it can be `undefined`

src/contexts/IEErrorContext/IEErrorContext.tsx
```diff
@@ -0,0 +1,20 @@
...
+const IEErrorContext = createContext(undefined);
+
+const IEErrorContextProvider = ({
+  children,
+  error,
+}: IEErrorContextProps): ReactElement => (
+  <IEErrorContext.Provider value={error}>{children}</IEErrorContext.Provider>
+);
+
+const useIEErrorContext = (): string => {
```
MkMan, line 14: Does IE stand for Internet Explorer?

src/components/Video/Video.tsx
```diff
@@ -81,27 +83,22 @@ const Video = forwardRef<HTMLDivElement, VideoProps>(function VideoRef(
 
   useEscapeKeyListener(stopVideo);
 
-  const captionsTrack = videoRef.current?.textTracks[0];
-
-  React.useEffect(() => {
+  useEffect(() => {
```
MkMan, line 86: This is nicer without `React.` Thanks 👍

src/components/Video/Video.tsx
```diff
@@ -81,27 +83,22 @@ const Video = forwardRef<HTMLDivElement, VideoProps>(function VideoRef(
...
     captionsTrack.mode = hasActiveCaptions ? 'showing' : 'hidden';
   }
 
-  /**
-   * @todo temporary fix for autoplay issue in ie -
-   * https://aesoponline.atlassian.net/browse/CON-426
-   * `const hasVideo = !!sizes`
-   */
-  const hasVideo = !!sizes && !(isIE && hasAutoplay && !isScrollBasedVideo);
+  console.log('ieError', ieError);
```
MkMan, line 99: ಠ_ಠ 
Is this temporary while testing? 😃

src/components/Video/Video.tsx
```diff
@@ -40,7 +41,8 @@ const Video = forwardRef<HTMLDivElement, VideoProps>(function VideoRef(
   },
   ref,
 ) {
-  const videoRef = useRef<HTMLVideoElement>();
+  const ieError = useIEErrorContext();
+  const videoRef = useRef<HTMLVideoElement>(null);
```
MkMan, line 45: This got me looking at the ref being passed into this component. Should it be assigned to the video element and used here?

src/components/Video/Video.tsx
```diff
@@ -81,27 +83,22 @@ const Video = forwardRef<HTMLDivElement, VideoProps>(function VideoRef(
...
 
-  const captionsTrack = videoRef.current?.textTracks[0];
-
-  React.useEffect(() => {
+  useEffect(() => {
     /** Stop and reset video if the source changes */
     if (hasMounted && videoRef.current) {
       videoRef.current.load();
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
MkMan, line 91: I am keen to get your thoughts on when we disable this rule and when we leave it with a warning? Currently, it gives me a warning in the IDE.

src/components/Video/Video.stories.mdx
beau-charman: I was testing til late last night - cleaning up the PR now

src/contexts/IEErrorContext/IEErrorContext.types.ts
beau-charman: that is true - was over engineering it

src/contexts/IEErrorContext/IEErrorContext.tsx
beau-charman: yes

src/components/Video/Video.tsx
beau-charman: 👍

src/components/Video/Video.tsx
beau-charman: yeah, cleaning up the PR now

src/components/Video/Video.tsx
beau-charman: Yeah, my IDE doesnt show me any of these warnings anymore now since the import of aesop-code-style :(

src/contexts/IEErrorContext/IEErrorContext.tsx
MkMan: Thanks for clarifying 🙂

The code doesn't appear to be specific to IE. Could this instead be generalised to `ErrorContext` to match the implementation?

beau-charman: Yeah I guess we could extend this later - the mindset was to was to have something temporary for IE support, but a generic Error context could come in handy later

src/components/Video/Video.tsx
MkMan: Dang, that's annoying 😕 ! I am guessing the Eslint plugin in Atom isn't playing nice with recursive configuration extensions.
Still, the output from the tool itself should be the ultimate source of truth. Should we bump up the severity of this rule to error so the build fails on commit (from lint staged) or on the CI?

beau-charman: that sounds fine, as in the limited situation where you dont want to check for updating dependancies, flagging it is a eslint ignore points out that it is intentional

MkMan (approved):

src/components/Video/Video.tsx
beau-charman: I will leave as is for now

PR #470, opened 2021-06-16
[Chore] Update readme and add component generator instructions

beau-charman:


README.md
```diff
@@ -34,7 +36,7 @@ npm install github:aesop/aesop-gel --save
 Import the component/hook/etc that you need in your code
 
 ```jsx
-import { Heading } from 'aesop-gel/dist/components/Heading';
+import { Heading } from 'aesop-gel';
```
MkMan, line 39: I am not fussed but that was intentional 😄 I thought new consumers of GEL should import just the components they need directly to avoid importing all of GEL #443

beau-charman: sorry! I thought this was left over from the beginning of (gel) time! I forgot that you committed that build change!

beau-charman: btw, love this. like how the lodash team provide their functions

MkMan (approved):

PR #490, opened 2021-06-29
[Feature/pur-324-upselltrack]: PDP Upsell functions

kevin-ho87:
Ticket
https://aesoponline.atlassian.net/browse/PUR-324

PDP Upsell expose functions for tracking when:
- scroll into view
- onClick


src/components/ProductDetailHeader/ProductDetailHeader.tsx
```diff
@@ -33,6 +35,8 @@ const ProductDetailHeader: ProductDetailHeaderType = ({
     onBreadcrumbClick(item, selectedVariant);
   const handleOnFlyinOpenCloseClick = (flyinState) =>
     onFlyinOpenCloseClick(flyinState, selectedVariant);
+  const handleOnUpsellScrollIntoView = () =>
+    onUpsellScrollIntoView && onUpsellScrollIntoView();
```
MkMan, line 39: Would this function return `false` if `handleOnUpsellScrollIntoView` isn't provided?

src/components/ProductDetailHeader/components/ProductDetailBody/ProductDetailBody.tsx
```diff
@@ -27,11 +28,25 @@ const ProductDetailBody: ProductDetailBodyType = ({
...
+    if (onUpsellScrollIntoView) {
+      onUpsellScrollIntoView();
+    }
+  };
+
+  useExecuteOnImpression(upsellRef, handleOnUpsellScrollIntoView, {
+    threshold: 0.2,
+    isExecutableOnReEntry: false,
+  });
+
```
zack76, line 49: how about 
```
if (onUpsellScrollIntoView) {
     useExecuteOnImpression(upsellRef, onUpsellScrollIntoView, {
     threshold: 0.2,
     isExecutableOnReEntry: false,
    });
}
```

zack76 (approved):

src/components/ProductDetailHeader/components/ProductDetailBody/ProductDetailBody.tsx
MkMan: Would that violate the [rule](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) on calling a hook inside a conditional?

zack76: ah sorry, i didnt realise that

src/components/ProductDetailHeader/ProductDetailHeader.tsx
kevin-ho87: I think it just returns `undefined` if `onUpsellScrollIntoView` prop is not provided

kevin-ho87:
```
const handleOnUpsellScrollIntoView = () =>
    !!onUpsellScrollIntoView && onUpsellScrollIntoView();
```
Would adding `!!` work better?

MkMan: I think what you already have is better 👍 `!!` would make `onUpsellScrollIntoView` into `false` if it was not provided and the function will then return `false`, right?

PR #501, opened 2021-07-22
TS Migration: CON-453 TextInputV2

MkMan:


src/components/TextInputV2/TextInputV2.spec.tsx
```diff
@@ -53,7 +53,7 @@ describe('<TextInputV2 />', () => {
     const onChangeFunction = jest.fn();
     render(
       <TextInputV2
-        isDisabled={true}
+        isEnabled={false}
```
hassanraza0054, line 56: Was this change simply a semantic one? Curious as I imagine "isEnabled" and "isDisabled" are both common and correct to use.

src/components/TextInputV2/TextInputV2.tsx
```diff
@@ -32,14 +32,14 @@ const TextInputV2 = forwardRef(
     const handleOnChange = useCallback(
       (event) => {
         setValue(event.target.value);
-        onChange(event);
+        onChange?.(event);
```
hassanraza0054, line 35: Very nice

src/components/TextInputV2/TextInputV2.tsx
```diff
@@ -32,14 +32,14 @@ const TextInputV2 = forwardRef(
...
       },
       [onChange, setValue],
     );
     const errorMessageId = `${inputId}-error-message`;
 
     return (
-      <div className={cx(styles.wrapper, classes.wrapper)}>
+      <div className={cx(styles.wrapper, classNames?.wrapper)}>
```
hassanraza0054, line 42: I don't think we've used optional chaining in **classNames** in other gel components have we? Thinking out loud, given we occasionally access it, might be a good general refactor rule to apply the same change you did here all across! 

Solid stuff

src/components/TextInputV2/TextInputV2.tsx
```diff
@@ -49,10 +49,10 @@ const TextInputV2 = forwardRef(
               styles.input,
               styles[theme],
               { [styles.hasError]: errorMessage },
-              classes.input,
+              classNames?.input,
             )}
             data-test-ref={dataTestRef}
-            disabled={isDisabled}
+            disabled={!isEnabled}
```
hassanraza0054, line 55: Ah is this why you changed the above property from "isDisabled" to "isEnabled"? If so, I see where you're coming from and agree with the change 👍

src/components/TextInputV2/TextInputV2.types.ts
```diff
@@ -0,0 +1,26 @@
+import type { ChangeEvent } from 'react';
+import type { Themes } from '~/types';
+
+export type TextInputV2Props = {
```
hassanraza0054, line 4: Just thinking out loud, in which cases will we not have optional parameters for a component?

src/components/TextInputV2/TextInputV2.tsx
hassanraza0054: Also might be a silly question, but in the case that we have the **styles** wrapper definition and **classNames** wrapper also provided, does one overwrite the other? Or does classNames add onto the already provided **styles** CSS logic? Curious what classNames is meant to be used for here from an implementation context standpoint

src/components/TextInputV2/TextInputV2.spec.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { TextInputV2 } from './';
+import { TextInputV2 } from '.';
```
zack76, line 4: how does this work? i've never tried this way of importing

MkMan: Same 😂 I didn't make that change myself, I think it's prettier or eslint that did it on commit 🤔 
It's the same as `'./index'`

src/components/TextInputV2/TextInputV2.spec.tsx
MkMan: This was to address [this](https://github.com/aesop/aesop-gel/blob/v1.9.0/src/components/TextInputV2/TextInputV2.js#L106) `TODO`. @beau-charman can probably explain it better.

src/components/TextInputV2/TextInputV2.tsx
MkMan: > I don't think we've used optional chaining in classNames in other gel components have we? Thinking out loud, given we occasionally access it, might be a good general refactor rule to apply the same change you did here all across!

I am not sure if there is a convention around this in the repo. If there is we can create GH issues to address them when a component is worked on.


> Also might be a silly question, but in the case that we have the styles wrapper definition and classNames wrapper also provided, does one overwrite the other? Or does classNames add onto the already provided styles CSS logic? Curious what classNames is meant to be used for here from an implementation context standpoint

The element will then have both classes applied to it, with the styles defined in later classes overriding those defined in earlier classes.

`classNames` is a way to provide class names for different html elements that make up `TextInputV2` to allow the user to make it look how they like.

src/components/TextInputV2/TextInputV2.types.ts
MkMan: I am not sure I understand the question 😕 To me, optional parameters are ones the component can function without them being provided. That could be because back up values are used.

If we take the `theme` prop as an example. As far as I am aware, it's optional in most components, that's because the components have a default value of `'dark'` and thus won't break and will still work if the prop hasn't got a value.

src/components/TextInputV2/TextInputV2.spec.tsx
beau-charman: boolean variable names are to be in the affirmative always. I am too old to figure out what the heck `notIsCouldBeDiabledButNotEnabled` is 😹

```
isEnabeld (or !!isEnabled)  // checking for the affirmative
!isEnabled // checking for the negative
```

src/components/TextInputV2/TextInputV2.spec.tsx
beau-charman: `.` means current directory, `./` is superfluous but stands out as for of a "directory" type string

hassanraza0054: Same Question!

src/components/TextInputV2/TextInputV2.types.ts
hassanraza0054: Ah so for example, if you look here at `SubNav.types.ts`, we haven't set **id** & **Links** as optional, which means the component must receive these, else it cannot function.

image


**id** exists in this component as well, but for now we're setting it as **optional**. So my questions are specifically around that choice. In particular:-

1. In what cases are we choosing to set types as **Optional** vs **Mandatory**?
2. What is the reasoning behind those choices?

More of a broad question for the team actually. Keen to hear your thoughts on this especially @beau-charman

hassanraza0054 (approved):
Approving for now as, besides my comments, ticket looks good to me!

PR #502, opened 2021-07-25
Chore/collaps dist diffs

MkMan:
image

This PR adds a `.gitattributes` file to collapse the diffs of our dist files as recommended in [this](https://medium.com/@clarkbw/managing-generated-files-in-github-1f1989c09dfd) article.

The files will still be there but at least collapsed to reduce clutter and space used in the PR. This is what they will look like 
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Generated Files
+dist/* linguist-generated=true


src/components/Button/Button.tsx
```diff
@@ -29,6 +29,7 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(function ButtonRef(
     styles[theme],
     { [styles.disabled]: !isEnabled },
     className,
+    'hahah REMOVE ME PLEASE'
```
zack76, line 32: 😕

MkMan: I left it intentionally for now so you can see the dist files collapsed in the files tab 😃

hassanraza0054: 👀

zack76 (approved):

src/components/Button/Button.tsx
MkMan: Commit now dropped from the PRs. Added a screenshot in the description to demo instead

PR #504, opened 2021-07-26
removed .vscode from gitignore

andrwchang:
Added plugin recommendation for VS Code - so the plugins mentioned in the read.me gets automatically suggested on load and the user can install all in one click


MkMan:
Which plugin did you install? 🙂 

I am slightly reluctant to version control the `.vscode` folder (or at least all of it). On the one hand IDE related files are not project files and should not be version controlled. On the other hand 90% of us use VS Code so 🤷‍♂️ 😃 

What I meant by adding the whole folder is the fact that it can contain user preferences, which in my opinion should not be forced on other people. For example, I have a plugin to change the colour of the IDE depending on the project open. The settings for it live inside the `.vscode` folder but if they get included in version: they're meaningless for people who don't use that plugin AND if they actually use it then I forcing my colour preference on them 😂

andrwchang:
> Which plugin did you install? 🙂
> 
> I am slightly reluctant to version control the `.vscode` folder (or at least all of it). On the one hand IDE related files are not project files and should not be version controlled. On the other hand 90% of us use VS Code so 🤷‍♂️ 😃
> 
> What I meant by adding the whole folder is the fact that it can contain user preferences, which in my opinion should not be forced on other people. For example, I have a plugin to change the colour of the IDE depending on the project open. The settings for it live inside the `.vscode` folder but if they get included in version: they're meaningless for people who don't use that plugin AND if they actually use it then I forcing my colour preference on them 😂

Installed typescript, babel, editorconfig, dbaeumer.vscode-eslint, prettier, stylelint.vscode-stylelint - so very basic plugins which I would expect everyone to use already. This was making the new starters life easier, however if majority of us have their own setup, best to ignore the `.vscode` folder. Would like to know what Others think

hassanraza0054:
Just having a quick look, I would personally definitely like having a standardised styling & code management settings file used for us all across different repos. 

That said, having had many discussions with devs across the team, the advice ive got is that it might be to unique a thing for everyone to have standardised. 

E.g. I use **custom prettier settings** and **format on save** because I like code falling neatly into place, even when im 
debugging and not committing. But I know some devs don't like prettier because it can make disruptive changes which are unintentional, and sometimes, **a small style change can make a file look like it was totally reworked**. I'm coming around stepping away from prettier slowly, but for now at least, its part of my workflow. 

To have a settings file knowing that its commonly used across the team for XYZ repo's is a Quality of life change i'd really appreciate. If concerns are that each repo has its own way of doing things, **we should document choices behind custom settings in one repo vs the other**. 

Those are my thoughts. Keen to hear more from others and hope my insights helped.

andrwchang:
Closing this as decided not to standardise the setup

PR #505, opened 2021-07-27
Ts/con 456 dynamic form

MkMan:


src/compositions/DynamicForm/DynamicForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+// };
+
+// DynamicForm.defaultProps = {
+//   children: undefined,
+//   className: undefined,
+//   defaultValues: undefined,
+//   formSchema: undefined,
+//   onSubmit: undefined,
+//   theme: undefined,
+// };
```
kevin-ho87, line 134: Should these comments be removed?

MkMan: Yes. Yes they definitely should be 😄  Thank you so much for picking up on that 🙏

src/compositions/DynamicForm/wrappers/TextField.tsx
```diff
@@ -1,9 +1,14 @@
-/* eslint-disable react/prop-types */
-import React from 'react';
+import React, { VFC } from 'react';
```
kevin-ho87, line 1: what is `VFC` ?

MkMan: Void Function Component, a Function Component that can't have `children`

https://www.mydatahack.com/using-react-vfc-instead-of-react-fc/

kevin-ho87: such a dark name. :P

kevin-ho87 (approved):

src/compositions/DynamicForm/DynamicForm.tsx
hassanraza0054: I wonder if we're keeping these now to see if any bugs come out with the Typescript conversion and we can revert back?

src/compositions/DynamicForm/wrappers/TextField.tsx
hassanraza0054: https://github.com/yannickcr/eslint-plugin-react/issues/2913

@kevin-ho87 from what i can tell, its a Type that allows you to define a component without the **children** prop. Correct me if im wrong or if there is more to it!

PR #506, opened 2021-07-27
feat: pass other props to button html

MkMan:
I was trying to assign `aria-expanded` to Button and found that nothing was happening. This change allows passing additional props to Button which will then be passed to the underlying <button>.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -17,6 +17,7 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(function ButtonRef(
     title,
     type = 'button',
     theme = 'dark',
+    ...otherProps
   },
   ref,
 ) {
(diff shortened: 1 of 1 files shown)


zack76 (approved):

beau-charman:
ah I with I got to this one. already doing work around aria roles, Button and the nav.

shouldn't just pass anything in as a `otherProps` wild card. 

for example, not all html elements take all aria attributes

beau-charman:
first diff file in my PR literally outlines this

MkMan:
> for example, not all html elements take all aria attributes
> first diff file in my PR literally outlines this

That's true but I felt that encompassing all possible attributes for each "low level" element would be difficult. For example, in the future if we wish to add `aria-label` that would require a PR. 
That's why I felt it's better to leave it open to the developer. Hints will be offered when using props specific to the component but beyond that they go to the dom.

beau-charman:
¯\_(ツ)_/¯

kevin-ho87:
Does this PR need to be merged into `main` instead of `develop` after the branch rename?

PR #508, opened 2021-07-27
feat: enhance Heading component

MkMan:
I realised I needed these changes when I tried to set focus to the heading but couldn't. These changes:

- Forward ref -> will enable passing a ref to call `.focus()` on
- Enable passing "other" attributes -> to be able to set `tabIndex={-1}`


kevin-ho87:
does `otherProps` need to be typed?

MkMan:
> does `otherProps` need to be typed?

What would it be typed as?

kevin-ho87:
I was thinking if it is specific to html props, something like this might work. Then again I have not tried it. If it already works finem just ignore me :P
```
const Input: React.FC<Props & React.HTMLProps<HTMLInputElement>> = props => { ... }
<Input about={...} accept={...} alt={...} ... />
```
https://github.com/piotrwitek/react-redux-typescript-guide#reacthtmlpropshtmlxxxelement

MkMan:
> I was thinking if it is specific to html props, something like this might work. Then again I have not tried it. If it already works fine just ignore me :P
> 
> ```
> const Input: React.FC<Props & React.HTMLProps<HTMLInputElement>> = props => { ... }
> <Input about={...} accept={...} alt={...} ... />
> ```
> 
> https://github.com/piotrwitek/react-redux-typescript-guide#reacthtmlpropshtmlxxxelement



I gave it a crack and got 2 errors 😞 

image

1. `Rest types may only be created from object types.` 
2. `Argument of type 'string' is not assignable to parameter of type 'Themes'.`

beau-charman:
why are we adding focus to heading elements?

MkMan:
> why are we adding focus to heading elements?

For collapsed elements, such as Accordions and Popups. When the element is expanded/enters the dom we need to let the user know.

My use case is the sub sections in the Account page.

image

When the user clicks Edit, I want to set the focus to the Heading, similar to [this](https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions#Example_1:_Using_a_button_to_collapse_and_expand_a_region) example.

beau-charman:
if its a focusable element and listens to user events, why not mark it up with a button?

MkMan:
> if its a focusable element and listens to user events, why not mark it up with a button?

I didn't understands what you meant by "listens to user events". But reasons for not placing it in a button are:
-  it's not clickable
- there is no intended interactivity

beau-charman:
> > if its a focusable element and listens to user events, why not mark it up with a button?
> 
> I didn't understands what you meant by "listens to user events". But reasons for not placing it in a button are:
> 
> * it's not clickable
> * there is no intended interactivity

"For collapsed elements, such as Accordions and Popups" sounds clickable to me

MkMan:
> > > if its a focusable element and listens to user events, why not mark it up with a button?
> > 
> > 
> > I didn't understands what you meant by "listens to user events". But reasons for not placing it in a button are:
> > 
> > * it's not clickable
> > * there is no intended interactivity
> 
> "For collapsed elements, such as Accordions and Popups" sounds clickable to me

The heading is not clickable, the Edit button is though 😃 You can see the behaviour yourself on `/my-account`

beau-charman:
i see what you mean, if its a pop up you can add something like 

```<div aria-hidden={!isActive} aria-label="Some description about the pop up. maybe have: (Press escape to close)" role="dialog" tabindex="0">```

to the wrapper

MkMan:
> i see what you mean, if its a pop up you can add something like
> 
> `<div aria-hidden={!isActive} aria-label="Some description about the pop up. maybe have: (Press escape to close)" role="dialog" tabindex={isActive ? 0 : null}>`
> 
> to the wrapper

I think that will do it but I have 2 concerns I wanted to discuss with you:
1. Every time we wish for a heading to be read and focused on we will need to wrap it in a div
2. This adds unnecessary div to the dom and I was under the impression that div soup was detrimental to screen reader

beau-charman:
> > i see what you mean, if its a pop up you can add something like
> > `<div aria-hidden={!isActive} aria-label="Some description about the pop up. maybe have: (Press escape to close)" role="dialog" tabindex={isActive ? 0 : null}>`
> > to the wrapper
> 
> I think that will do it but I have 2 concerns I wanted to discuss with you:
> 
> 1. Every time we wish for a heading to be read and focused on we will need to wrap it in a div
> 2. This adds unnecessary div to the dom and I was under the impression that div soup was detrimental to screen reader

headings will be read either way, they don't need a tab index, or focus. I am assuming that you are wrapping to popup or accordion in some markup - put the attributes in there

hassanraza0054 (approved):
Will just tag @beau-charman to confirm if `otherProps` is alright to go ahead with name wise. Other than that, solid accessibility change!

hassanraza0054:
> you

What would be a suitable markup in this case? Curious to get your thoughts @beau-charman

hassanraza0054:
> > > i see what you mean, if its a pop up you can add something like
> > > `<div aria-hidden={!isActive} aria-label="Some description about the pop up. maybe have: (Press escape to close)" role="dialog" tabindex={isActive ? 0 : null}>`
> > > to the wrapper
> > 
> > 
> > I think that will do it but I have 2 concerns I wanted to discuss with you:
> > 
> > 1. Every time we wish for a heading to be read and focused on we will need to wrap it in a div
> > 2. This adds unnecessary div to the dom and I was under the impression that div soup was detrimental to screen reader
> 
> headings will be read either way, they don't need a tab index, or focus. I am assuming that you are wrapping the popup or accordion in some markup - put the attributes in there

What would be a suitable markup in this case? Curious to get your thoughts @beau-charman

MkMan:
After catching up with @beau-charman this may not be needed, closing the PR.

PR #510, opened 2021-07-28
Form field enhancements

MkMan:
- fix `onChange` TS type on InputTextV2
- Improvements to DynamicForm
  - Add validation rule to check 1 field equals another
  - Allow specifying validation behaviour (onChange, onSubmit, etc)
  - Fix onSubmit TS type

### DIFF
--- a/src/compositions/DynamicForm/DynamicForm.types.ts
+++ b/src/compositions/DynamicForm/DynamicForm.types.ts
@@ -1,4 +1,5 @@
-import type { ReactNode, SyntheticEvent } from 'react';
+import type { ReactNode } from 'react';
+import type { Mode } from 'react-hook-form/dist/index.ie11';
 import type { SelectProps } from '~/components/Select/Select.types';
 import type { TextInputV2Props } from '~/components/TextInputV2/TextInputV2.types';
 import type { Themes } from '~/types';
@@ -47,6 +48,7 @@ export type DynamicFormProps = {
   className?: string;
   defaultValues?: Record<string, string>;
   formSchema?: FormFieldsRow[];
-  onSubmit: (event: SyntheticEvent<HTMLFormElement>) => void;
+  onSubmit: (formData: Record<string, string>) => void;
   theme?: Themes;
+  validationMode?: Mode;
 };
--- a/src/compositions/DynamicForm/validators/validators.spec.ts
+++ b/src/compositions/DynamicForm/validators/validators.spec.ts
@@ -2,6 +2,8 @@ import { getValidationRules } from './validators';
 import type { AvailableFormFieldTypes } from '../wrappers';
 
 describe('DynamicForm field validators', () => {
+  const getValuesMock = jest.fn().mockReturnValue('Test1234');
+
   describe(`TextField validation rules`, () => {
     const fieldType = 'TextField';
 
@@ -19,9 +21,9 @@ describe('DynamicForm field validators', () => {
         },
       };
 
-      expect(getValidationRules(schemaRequirements, fieldType)).toEqual(
-        expectedReturn,
-      );
+      expect(
+        getValidationRules(schemaRequirements, fieldType, getValuesMock),
+      ).toEqual(expectedReturn);
     });
 
     it('should assign the maximum length validation object from the schema', () => {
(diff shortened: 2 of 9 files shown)


src/compositions/DynamicForm/DynamicForm.types.ts
```diff
@@ -47,6 +48,7 @@ export type DynamicFormProps = {
   className?: string;
   defaultValues?: Record<string, string>;
   formSchema?: FormFieldsRow[];
-  onSubmit: (event: SyntheticEvent<HTMLFormElement>) => void;
+  onSubmit: (formData: Record<string, string>) => void;
```
hassanraza0054, line 51: Ooh what's this Record type and whats the value/benefit of the change? Curious!

src/compositions/DynamicForm/validators/validators.spec.ts
```diff
@@ -2,6 +2,8 @@ import { getValidationRules } from './validators';
 import type { AvailableFormFieldTypes } from '../wrappers';
 
 describe('DynamicForm field validators', () => {
+  const getValuesMock = jest.fn().mockReturnValue('Test1234');
```
hassanraza0054, line 5: Lost opportunity for "Kek". Classic.

src/compositions/DynamicForm/validators/validators.ts
```diff
@@ -1,9 +1,11 @@
+import type { UseFormMethods } from 'react-hook-form/dist/index.ie11';
```
hassanraza0054, line 1: This looks new. Whats this for?

src/compositions/DynamicForm/validators/validators.ts
```diff
@@ -26,6 +28,13 @@ export const getValidationRules = (
         rules.maxLength = validationObject.maxLength;
       }
 
+      // Text equals another field
+      if (validationObject.isSameAs) {
+        rules.validate = (value: string) =>
```
hassanraza0054, line 33: Ah so if this rule exists, we're constructing the value the rule should check the *sameness* against? And thus the message that should accompany when a match occurs. And all this returns to be used by the form. 

That flow correct?

src/compositions/DynamicForm/DynamicForm.types.ts
```diff
@@ -47,6 +48,7 @@ export type DynamicFormProps = {
...
   theme?: Themes;
+  validationMode?: Mode;
```
zack76, line 53: i'm just curious how you knew there was a `Mode` type can be used here

src/compositions/DynamicForm/DynamicForm.types.ts
MkMan: I originally typed the `onSubmit` function incorrectly; marking it as taking an event type argument when in fact it receives an object of form field names and their values.

Attempting to use it prior to the change would be misleading for developers and would require overriding TS to correctly use the function.

`Record<string, string>` is an object whose keys are strings and values are strings.

src/compositions/DynamicForm/validators/validators.ts
MkMan: Instead of manually specifying the type of an argument which comes from the library in their first place, I used the type exported by the library.

src/compositions/DynamicForm/DynamicForm.types.ts
MkMan: `CMD + click` on the `mode` field of `useForm` took me to its TS definition, from there I traced down `Mode` 😃

src/compositions/DynamicForm/validators/validators.ts
MkMan: Not quite 🙂 The validation rules object can have a `validate` function that takes the current field value and as an argument. If it returns `true` the field is considered valid. If it returns a string, the field is considered invalid and that string is the validation error message.

src/compositions/DynamicForm/DynamicForm.types.ts
hassanraza0054: Got it! Nice

src/compositions/DynamicForm/validators/validators.ts
hassanraza0054: Gotcha!

hassanraza0054 (approved):

PR #527, opened 2021-08-09
[Feature/pur 446 copy]: ATC label

kevin-ho87:
Ticket:
https://aesoponline.atlassian.net/browse/PUR-446

Description:
Hide `-` on Add to Cart label if text is not provided

image


zack76 (approved):

src/components/AddToCartButton/AddToCartButton.spec.tsx
```diff
@@ -73,6 +73,67 @@ describe('<AddToCartButton />', () => {
...
+        product={product}
+        variants={variants}
+      />,
+    );
+
+    const button = screen.getByRole('button', {
+      name: `${fixture.variants[0].price}`,
+    });
+
+    expect(button).toBeInTheDocument();
```
hassanraza0054, line 93: Found it a bit confusing as to why the button should render a label without cartAction. So had a look in the code and fixture to verify. Apparently the title is either **cartActionLabel** or **copy?.outOfStock?.title** with the latter being part of the fixture and thus, what's allowing a label render.

Just curious and I could be wrong. Would it be better to write 'should render label as default out of stock without cartAction copy' ?
image

hassanraza0054: As an alternative, since we're checking the label itself, would it be clearer to perhaps keep the test name the same, but mock a **Default Title** and test not only that the button has rendered, but also that said rendered label = **Default Title**? That way, anyone going over these tests can instantly understand what is considered the default label.

src/components/AddToCartButton/AddToCartButton.spec.tsx
```diff
@@ -73,6 +73,67 @@ describe('<AddToCartButton />', () => {
...
+        product={product}
+        variants={variants}
+      />,
+    );
+
+    const button = screen.getByRole('button', {
+      name: `${fixture.variants[0].price} ${fixture.copyJP.postTaxPrice}`,
+    });
+
+    expect(button).toBeInTheDocument();
```
hassanraza0054, line 134: Yeah this one im a bit curious of as well. Are we only checking if the button exists with the props provided? Or do we also want to confirm the label? Because here, the label would be **postTaxPriceText** instead of **copy?.outOfStock?.title**. But its not as obvious in the assertion of this test to me. I could be wrong and overthinking this though. Keen to hear your thoughts.

src/components/AddToCartButton/AddToCartButton.tsx
```diff
@@ -38,9 +38,12 @@ const AddToCartButton: AddToCartButtonType = ({
...
     sku,
   } = selectedVariant;
-  const cartActionLabel = !copy?.postTaxPrice
+
+  const cartActionText = copy?.cartAction
     ? `${copy?.cartAction} — ${price}`
-    : `${copy?.cartAction} — ${price} ${copy?.postTaxPrice}`;
+    : price;
+  const postTaxPriceText = copy?.postTaxPrice ? ` ${copy?.postTaxPrice}` : '';
+  const cartActionLabel = cartActionText + postTaxPriceText;
```
hassanraza0054, line 46: What would a label look like if **copy?.cartAction** exists as well as **copy?.postTaxPrice**?

In English, im imagining '**Add to your cart - $30 ($35)**' looking at the fixture and comparing the JP and EN copy.

src/components/AddToCartButton/AddToCartButton.spec.tsx
kevin-ho87: This for a scenario if an empty string gets passed into the `cartActionLabel`, eg: an error occurs in the app consuming it. Example: it renders `$100` instead of ` - $100`. Basically remove the dash.

src/components/AddToCartButton/AddToCartButton.spec.tsx
kevin-ho87: Both, to check button exists and if label is correctly passed/read.

src/components/AddToCartButton/AddToCartButton.tsx
kevin-ho87: `Add to your cart - $30 (after tax tax)`

src/components/AddToCartButton/AddToCartButton.spec.tsx
hassanraza0054: Ah. Just to confirm, by **this**, do you mean the test case i've commented on above, or the screenshot of the 2 lines of code I shared?

src/components/AddToCartButton/AddToCartButton.spec.tsx
hassanraza0054: Alright so the label we're expecting to read here is **postTaxPriceText** if i'm not mistaken?

src/components/AddToCartButton/AddToCartButton.tsx
hassanraza0054: Thanks for that! Learning a lot from this

src/components/AddToCartButton/AddToCartButton.spec.tsx
kevin-ho87: price + postTaxPrice

src/components/AddToCartButton/AddToCartButton.spec.tsx
kevin-ho87: This ticket in general

hassanraza0054 (approved):

MkMan (approved):
You could probably combine your test cases into 1 `test.each` seeing that the test implementation is almost identical across them.

PR #528, opened 2021-08-10
[Feature/CON-128]: add ref for impression tracking

hassanraza0054:
Ticket: https://aesoponline.atlassian.net/browse/CON-128

Related web-ui ticket: https://github.com/aesop/aesop-web-ui/pull/3991

### DIFF
--- a/src/components/DoubleMedia/DoubleMedia.tsx
+++ b/src/components/DoubleMedia/DoubleMedia.tsx
@@ -1,23 +1,20 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import cx from 'classnames';
 import { MediaBlock } from './components/MediaBlock';
-import type { DoubleMediaType } from './DoubleMedia.types';
+import type { DoubleMediaProps } from './DoubleMedia.types';
 import styles from './DoubleMedia.module.css';
 
-const DoubleMedia: DoubleMediaType = ({
-  className,
-  mediaOne,
-  mediaTwo,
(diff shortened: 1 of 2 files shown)


MkMan (approved):

src/components/DoubleMedia/DoubleMedia.tsx
```diff
@@ -1,23 +1,20 @@
...
 
-const DoubleMedia: DoubleMediaType = ({
-  className,
-  mediaOne,
-  mediaTwo,
-  theme = 'dark',
-}) => {
-  const classSet = cx(styles.base, styles[theme], className);
+const DoubleMedia = forwardRef<HTMLDivElement, DoubleMediaProps>(
+  ({ className, mediaOne, mediaTwo, theme = 'dark' }, ref) => {
```
zack76, line 8: the format looks a bit weird. Have you run `npm run format`

zack76 (approved):

src/components/DoubleMedia/DoubleMedia.tsx
MkMan: Prettier sometimes formats it that way (like [here](https://github.com/aesop/aesop-gel/blob/de05a279afce01754055a9264f31bf84655e400e/src/components/Accordion/Accordion.tsx#L17)). I think it's because the line isn't longer than threshold so it's not breaking it up to 1 prop per line.

PR #529, opened 2021-08-12
Create personal account info summary comp

MkMan:
The purpose is to move out [this](https://github.com/aesop/aesop-web-ui/blob/develop/src/app/components-custom/account/information/AccountInfoPersonal.js#L49-L80) component so it can be used in aesop-com.

#### Todo
- [x] Tests
- [x] Export the component so it can be consumed


zack76 (approved):
Awesome!

PR #533, opened 2021-08-20
[RET-251]Accessible loading state

zack76:
Adding the screen reader only text to the loading indicator so that the loading state can be announced


kevin-ho87 (approved):

MkMan (approved):
I am not an expert with accessibility so I can't offer feedback, but we can try this and refine/fix based on the results 👍

PR #534, opened 2021-08-20
Feature [CON-625] Global Navigation Context

beau-charman:
To address https://aesoponline.atlassian.net/browse/CON-625, this PR refactors the Mobile Navigation components and context to be consumed within a Global Navigation context (and optional wrapper component)

 https://aesoponline.atlassian.net/browse/CON-625


maxlew (approved):
It all makes sense to me, I think it's a lot cleaner now that it sits under the one GlobalNavigation. 

My only questions were if Tablet and Desktop would share a context? And if there's any downside to creating/destroying the context providers on a resize?

beau-charman:
> It all makes sense to me, I think it's a lot cleaner now that it sits under the one GlobalNavigation.
> 
> My only questions were if Tablet and Desktop would share a context? And if there's any downside to creating/destroying the context providers on a resize?

Really good point, if there were any event handlers or asyc effects or async state updates - they would need to be cleaned up. I am however trying to keep the state super simple. If however - state needs to persist across view changes - those states can be lifted to a higher context

beau-charman:
> It all makes sense to me, I think it's a lot cleaner now that it sits under the one GlobalNavigation.
> 
> My only questions were if Tablet and Desktop would share a context? And if there's any downside to creating/destroying the context providers on a resize?

Also, Tablet and Desktop will get there own context if need, and share the two exported from `GlobalNavigation`

PR #535, opened 2021-08-24
[Global Navigation] Update exports for easier consumption

maxlew:
I'm not 100% sure if this is the best way to go about this, but in the consuming application we need to access the Component, Types, and Context in a single place - example below. 

```tsx
import React from 'react';
import {
  GlobalNavigation,
  GlobalNavigationType,
  GlobalNavigationContextProvider,
  GlobalNavigationStateContextProvider
} from '@aesop-engineering/gel/dist/components/GlobalNavigation';

// TODO replace with graphql calls
import { MobileNavigationFixture as data } from './navigation.fixture';

type ActionHandlers = {
  search?: () => void;
  cart?: () => void;
  read?: () => void;
  account?: () => void;
  visit?: () => void;
  support?: () => void;
}

type NavigationProps = {
  isVisuallyObstructed: boolean;
  actionHandlers: ActionHandlers;
};

export const Navigation = (
  {
    isVisuallyObstructed = false,
    actionHandlers = {}
  }: NavigationProps
): React.ReactElement => {

  // TODO Loading State?

  // Combine actions with actionHandlers from parent
  const actions = Object.keys(data.actions).reduce((accActions, action) => {
    accActions[action] = data.actions[action];

    // adding 'onClick' handler for triggers when a match is found in the handlers
    if (accActions[action].type === 'trigger' && actionHandlers[action]) {
      accActions[action].onClick = actionHandlers[action];
    }

    return accActions;
  }, {});

  const NavigationProperties = {
    actions,
    articles: data.articles,
    collections: data.collections,
    isVisuallyObstructed,
    mobileViewClosedTheme: 'light',
    theme: 'light',
  }  as GlobalNavigationType.GlobalNavigationContextProviderType

  return (
    <GlobalNavigationStateContextProvider>
      <GlobalNavigationContextProvider value={NavigationProperties}>
        <GlobalNavigation />
      </GlobalNavigationContextProvider>
    </GlobalNavigationStateContextProvider>
  );
};

export default Navigation;
```


MkMan (approved):
Looks good to me, it would be good to get Beau's thoughts on it as well :)

PR #542, opened 2021-09-07
[Feature/RET-317] Reorder Cta

zack76:
Update the border color and text color of active/inactive buttons


MkMan:
Is the UX team across this change? It will impact the whole website 😃

zack76:
> Is the UX team across this change? It will impact the whole website 😃

yes, it's a change requested by Liz and James also approved it.
The change is supposed to be applied to all gel buttons

MkMan (approved):

PR #546, opened 2021-09-10
feat: updated the readme to make the link work with the correct npm r…

andrwchang:
running `npm link aesop-gel` from aesop-web-ui does not work as there is no npm registry exists with that name.
updated to point to the correct npm registry


zack76:
I think this one hasn't been merged yet
https://github.com/aesop/aesop-web-ui/pull/4024

andrwchang:
@zack76 This can go in as this is only the change in readme to fix the issue in dev.

zack76 (approved):

hassanraza0054:
Keen to get thoughts on this PR @maxlew 

As @zack76 mentioned, can we drop this in favour of merging aesop/aesop-web-ui#4024 and documenting the new process in the readme instead?

MkMan:
Is this PR still needed, @andrwchang ?

PR #550, opened 2021-09-14
feat: fix active panel issue

beau-charman:


src/components/GlobalNavigation/components/DesktopView/components/Panel/Panel.module.css
```diff
@@ -15,6 +15,10 @@
   pointer-events: none;
   transition: border-top-color 200ms var(--easing-ease-in-out-cubic);
 
+  &.active {
```
hassanraza0054, line 18: So just to get some clarity, what is the splitting of the active class with z-index in one and opacity in the other (the latter nested in content and former at base) fixing?

hassanraza0054 (approved):

src/components/GlobalNavigation/components/DesktopView/components/Panel/Panel.module.css
beau-charman: I want the panels and background colours to be always visible so that the transitions are nicer, so the opacity and pointer events of the inner containers are what really changes

PR #553, opened 2021-09-14
[Feature/Nav]: Add dataTestRefs for desktop view

hassanraza0054:
As per QA requirements, added data-test-refs for highlighted regions in new Nav for Desktop view.

https://lucid.app/lucidchart/invitations/accept/inv_9be02fa5-a747-4e4d-a0a1-95f90395dc32

**Changes Video**
https://user-images.githubusercontent.com/42990483/133234043-646f46f5-daad-4f14-bda6-e9d756fe27e2.mov

**Identifiers used**
TLC => Top Level Collection
NECO => Nested Collection
NONECO => Notable Nested Collection
TOD => Taxonomy Of Design

With **Read**, I chose to identify the second ul, i.e. the **Read-Collection**, by adding the suffix "-READ" ahead of the TLC naming. This is to accomodate the implementation in code, where its seen that a second ul only comes in the first column of the nav panel if it is a **Read-Collection** (Screenshot below for reference).

image

Keen to get feedback. Chose data-test-ref names with the following assumptions:-

1. ID's will remain most consistent throughout development of the nav, making them great candidates to construct the data test ref strings.
2. Naming scheme (i.e. acronyms/identifiers chosen above) are self-explanatory based on the understanding of first column, second column and third column items in the Nav hosting **Top Level Collections**, **Nested Collections** & **Notable Nested Collections** respectively.

### DIFF
--- a/src/components/GlobalNavigation/components/DesktopView/components/CollectionImage/CollectionImage.tsx
+++ b/src/components/GlobalNavigation/components/DesktopView/components/CollectionImage/CollectionImage.tsx
@@ -3,11 +3,11 @@ import { Image } from '~/components/Image';
 import type { CollectionImageType } from './CollectionImage.types';
 import styles from './CollectionImage.module.css';
 
-const CollectionImage: CollectionImageType = ({ altText, sizes }) => {
+const CollectionImage: CollectionImageType = ({ altText, sizes, dataTestRef }) => {
   if (!sizes) return null;
 
   return (
-    <div className={styles.base}>
+    <div className={styles.base} data-test-ref={dataTestRef}>
       <Image
         altText={altText}
         className={styles.image}
--- a/src/components/GlobalNavigation/components/DesktopView/components/CollectionImage/CollectionImage.types.ts
+++ b/src/components/GlobalNavigation/components/DesktopView/components/CollectionImage/CollectionImage.types.ts
@@ -2,6 +2,7 @@ import type { ComponentWithoutChildren } from '~/types';
 
 type CollectionImageProps = {
   altText?: string;
+  dataTestRef?: string;
   sizes?: {
     medium?: string;
     large?: string;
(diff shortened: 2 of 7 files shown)


MkMan:
What's the reason behind passing these test ref values in as props instead of hard coding them in the component (by referencing a constants file)? Especially, that they don't get passed in from the consumer of the navigation component?

beau-charman:
> What's the reason behind passing these test ref values in as props instead of hard coding them in the component (by referencing a constants file)? Especially, that they don't get passed in from the consumer of the navigation component?

Because, taking the screen grab for example the `currentId` will reference the current collection in the iteration, 

For example `NAV_SKIN-CARE_TLC`, `NAV_BODY_TLC`

hassanraza0054:
> What's the reason behind passing these test ref values in as props instead of hard coding them in the component (by referencing a constants file)? Especially, that they don't get passed in from the consumer of the navigation component?

Good question. At the moment, we don't have a standard of maintaining **dataTestRef** constants in our gel. I sense that part of that is because of the different ways in which a **dataTestRef** can be constructed, thus not always making it a true constant.

E.g. below is an example of a prop being passed. All's well and good, no **dataTestRef** constant required.
image

But in a few other places, the **dataTestRefs** are dependent partly on certain data, like ID's here for example.

image

Ofcourse, we also have an example of a completely hardcoded string fallback in case the value isnt.

So those are the reasons I wasn't completely confident in implementing a constants file, since there wasn't a **1 to 1** transformation.

Finally, regarding your question on **why props over hardcoding in components**, the **CollectionLayout** component is the higher order component that has the id necessary to help create a distinguishable dataTestRef. Let me explain. 

image

Each expanded desktop panel has its own unique set of collections that need to be tested with the help of distinguishable dataTestRefs. Now with our current implementation, while as you can see, skin care is open in the image below, in actuality, the collections that fall under the panels of the other menu items also simultaneously exist on the DOM. They're just hidden by CSS & JS. That means there is a dependency to make distinct dataTestRefs based on a higher level, stable string. And in this context, what I found that to be was the **currentId** value, which is managed neatly in CollectionLayout.js.

image

image

MkMan (approved):

beau-charman (approved):

zack76 (approved):

PR #556, opened 2021-09-15
[Feature/CON-626]: Add mobile navigation test cases 

hassanraza0054:
Add comprehensive testing to the mobile navigation testing. 

Since there was a lot of code & changes, and I was trying to cover cases I thought were most fruitful for future devs working on this, there are probably are fair number of areas the team can give me suggestions or input on how I wrote these tests. 

So while this is near completion, im hoping to leverage everyones thoughts before pushing this across the line. Also the code does feel fairly wet in some areas. Keen to get tips on optimising.

Also added an optional chaining condition to the header. Found out that optimisation while writing test cases. TDD at its finest 👌


MkMan:
Can you please exlcude `package-lock.json` from your changes? It shouldn't change unless there are accompanying `package.json` changes.

src/components/GlobalNavigation/components/MobileView/components/Collection/Collection.spec.tsx
```diff
@@ -1,8 +1,148 @@
...
+    const nestedCollectionButton = screen.getByRole('button');
+    nestedCollectionButton.click();
+
+    expect(setActiveNestedCollectionIds).toHaveBeenCalledTimes(1);
+  });
+
+  it('should render promotion list item', () => {
+    render(<TestBed promotion={CollectionFixture.promotion} />);
+
+    expect(listItemSpy).toHaveBeenCalledTimes(2);
```
MkMan, line 146: If possible use something more dynamic than `2`. It makes the intention clearer and is less brittle (if more items are added).

MkMan (commented):
Good stuff man 👏 

I went through about half of the changes locally and thought I'd make some suggestions based on patterns I saw.
- Can you please exclude `package-lock.json` from your changes? It shouldn't change unless there are accompanying `package.json` changes.
- With regards to Snapshot testing, what do you think about the discussion in [this](https://aesop.slack.com/archives/C0106S5K63W/p1619732664084600) thread?
- Please try to rely on 'All' type queries and accessing elements by index. If the DOM structure changes the tests fail when what they're testing may actually still be working. It's better to be more specific with the queries and not let your knowledge of the dom structure influence the query.

src/components/GlobalNavigation/__mocks__/GlobalNavigation.context.tsx
```diff
@@ -0,0 +1,39 @@
...
+  setActiveCollectionId,
+});
+
+const GlobalNavigationContextProvider = ({ children }) => <>{children}</>;
+
+const useGlobalNavigationContext = jest.fn().mockReturnValue({
+  actions,
+  onOpen,
+  activeView,
+}); // TODO: Set default value here later
```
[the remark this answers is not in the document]
hassanraza0054: Yup this is left for later. Not relevant atm. Will remove after I complete the desktop navigation tests which uses this branch

src/components/GlobalNavigation/components/MobileView/components/ArticleCard/ArticleCard.spec.tsx
```diff
@@ -1,8 +1,57 @@
...
+          url={ArticleCardFixture.url}
+        />
+      </>,
+    );
+
+    const focusableElements = screen.getAllByRole('link');
+    const one = focusableElements[0];
+
+    expect(one).not.toHaveAttribute('tabIndex');
+  });
```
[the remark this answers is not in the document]
hassanraza0054: Ah this was writing the tab click tests early on. I think later, I changed the test logic to be "xyz is tabbable" which just checks the tabIndex status of the element in question. Great catch, I'll change this test name to something like "should allow article card to be tabbable".

src/components/GlobalNavigation/components/MobileView/components/Collection/Collection.spec.tsx
```diff
@@ -1,8 +1,148 @@
...
+    const { setActiveCollectionId } = useGlobalNavigationStateContext();
+
+    const buttons = screen.getAllByRole('button');
+    const topCollectionButton = buttons[0];
+    topCollectionButton.click();
+
+    expect(setActiveCollectionId).toHaveBeenCalledTimes(1);
+  });
+
+  it('should have tabbable top collection button if visible', () => {
```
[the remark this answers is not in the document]
hassanraza0054: Yup its correct in this situation as the default of 'an element being tabbable' is when there is a tabIndex of **null**, which means when rendered in the dom that it doesnt exist, thus allowing the element to be tabbable. The alternative is that tabIndex would be '-1', making it render as a dom attribute and making its respective element inaccessible.

src/components/GlobalNavigation/components/MobileView/components/ArticleCard/ArticleCard.spec.tsx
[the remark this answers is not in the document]
hassanraza0054: That said, I've added an extra test explicitly testing tabs. Thanks!

src/components/GlobalNavigation/components/MobileView/components/Collection/Collection.spec.tsx
```diff
@@ -1,8 +1,148 @@
...
+    );
+  };
+
+  it('should render base component correctly', () => {
+    const { container } = render(<TestBed />);
+
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should not render top collection', () => {
```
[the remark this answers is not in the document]
hassanraza0054: Very valid point. Added!

src/components/GlobalNavigation/components/MobileView/components/Collection/Collection.spec.tsx
```diff
@@ -1,8 +1,148 @@
...
+    render(<TestBed shouldHideTopCollection={true} />);
+    const buttons = screen.getAllByRole('button');
+
+    expect(buttons.length).toEqual(1);
+    expect(buttons[0]).not.toHaveAttribute('title', CollectionFixture.title);
```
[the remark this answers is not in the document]
hassanraza0054: So in this case, there can only be 2 buttons. The top button, which doesnt get rendered by the boolean `shouldHideTopCollection` being set to true, and the back button, which still renders.

Unfortunately since there is no 'name' to query by, I cant directly call said button without the use of titles (something both buttons do have). Plus these buttons aren't dependent on a list being provided. They always render, with the top button only not rendering if the boolean is set to false.

So I assumed that this test was good, as if it does fail in the future due to another button being added, that gives us valuable input to change our test as a significant functionality change would've probably occurred.

src/components/GlobalNavigation/components/MobileView/components/Collection/Collection.spec.tsx
```diff
@@ -1,8 +1,148 @@
...
+    (useGlobalNavigationStateContext as jest.Mock).mockReturnValue({
+      activeCollectionId: 'top',
+      setActiveCollectionId: jest.fn(),
+    });
+
+    const { activeCollectionId } = useGlobalNavigationStateContext();
+
+    render(<TestBed id={activeCollectionId} />);
+
+    const { setActiveCollectionId } = useGlobalNavigationStateContext();
```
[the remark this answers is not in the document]
hassanraza0054: Good catch. No reason, probably just missed that.

src/components/GlobalNavigation/components/MobileView/components/Collection/Collection.spec.tsx
```diff
@@ -1,8 +1,148 @@
...
+  it('should have tabbable top collection button if visible', () => {
+    (useGlobalNavigationStateContext as jest.Mock).mockReturnValue({
+      activeCollectionId: 'top',
+      setActiveCollectionId: jest.fn(),
+    });
+
+    const { activeCollectionId } = useGlobalNavigationStateContext();
+    render(<TestBed id={activeCollectionId} isVisible={true} />);
+
+    const topCollectionButton = screen.getAllByRole('button')[0];
```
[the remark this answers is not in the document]
hassanraza0054: Changed this one and the above to getByTitle instead!

src/components/GlobalNavigation/components/MobileView/components/ArticleCard/ArticleCard.spec.tsx
```diff
@@ -1,8 +1,49 @@
...
+
+    expect(container).not.toBeEmptyDOMElement();
+  });
+
+  it('should navigate to the article card by tab clicking', () => {
+    render(
+      <>
+        <TestBed />
+      </>,
+    );
```
MkMan, line 29: I don't think you need to wrap it in a fragment, right?
```suggestion
    render(<TestBed />);
```

src/components/GlobalNavigation/components/MobileView/components/ArticleCard/ArticleCard.spec.tsx
```diff
@@ -1,8 +1,49 @@
...
+
+    const focusableElements = screen.getAllByRole('link');
+    const one = focusableElements[0];
```
MkMan, line 32: Just a couple of things here:
1. technically, `focusableElements` is a list of links. Button is focusable but it wouldn't be included in this array
2. `one` could probably be better named, perhaps `firstLink`?

The same applies to the following `it`.

src/components/GlobalNavigation/components/MobileView/components/ArticleCard/ArticleCard.spec.tsx
```diff
@@ -1,8 +1,49 @@
...
+
+    userEvent.tab();
```
MkMan, line 34: Nice 👌 👏

src/components/GlobalNavigation/components/MobileView/components/ArticleList/ArticleList.spec.tsx
```diff
@@ -1,7 +1,81 @@
-// import React from 'react';
-// import { render, screen } from '@testing-library/react';
-// import { ArticleList } from './ArticleList';
+/* eslint-disable init-declarations */
```
MkMan, line 1: You should be able to remove these from all your files now that #559 is merged 😃

src/components/GlobalNavigation/components/MobileView/components/Header/Header.spec.tsx
```diff
@@ -1,8 +1,237 @@
...
+    (useGlobalNavigationContext as jest.Mock).mockReturnValue({
+      ...testGlobalNavigationContext,
+    });
+    (useGlobalNavigationStateContext as jest.Mock).mockReturnValue({
+      ...testGlobalNavigationStateContext,
+    });
+    (useMobileViewContext as jest.Mock).mockReturnValue({
+      ...testMobileViewContext,
+    });
+    logoSpy = jest.spyOn(logoFile, 'Logo').mockImplementation(() => null);
```
MkMan, line 64: IDE is saying `logoSpy` is not being accessed. You can just do the `jest.spyOn(...` without assigning its return.

src/components/GlobalNavigation/components/MobileView/components/Header/Header.spec.tsx
```diff
@@ -1,8 +1,237 @@
...
+    });
+    const { onOpen } = useGlobalNavigationContext();
+    const { setIsOpen } = useGlobalNavigationStateContext();
+
+    render(<TestBed />);
+
+    const closeBtn = screen.getAllByRole('button')[2];
+    closeBtn.click();
+
+    expect(onOpen).toEqual(null);
```
MkMan, line 157: Wait ... Aren't you setting it to null on like 147 😅 ?

hassanraza0054: Yeah its one of those cases where its just making doubly sure that the situation is indeed being **Mocked**. The important assertion here is infact the one beneath, i.e. expect(setIsOpen).toHaveBeenCalledTimes(1).

I just figured since it's good to add as much context to the mock test as possible, might as well keep this fairly obvious line 157 check here. After all, if I mocked it incorrectly, the value on line 149 could indeed have been different.

Open to changing this though as I know I have a lot to learn.👌

src/components/GlobalNavigation/components/MobileView/components/Header/Header.tsx
```diff
@@ -77,6 +77,7 @@ const Header: HeaderType = ({ onClose }) => {
           <li className={styles.item}>
             <Button
               className={cx(styles.action, styles.actionSearch)}
+              data-testid={'NAV_SEARCH'}
```
MkMan, line 80: Is it possible to make the test ids/refs in this component a bit more descriptive? They currently don't indicate that they're on buttons but rather menus.

src/components/GlobalNavigation/components/MobileView/components/Link/Link.spec.tsx
```diff
@@ -1,8 +1,26 @@
...
+    const link = screen.getByRole('link');
+    expect(link).not.toHaveAttribute('tabIndex');
+  });
+
+  it('should be an external link', () => {
+    render(<Link {...LinkFixture} isExternal={true} />);
+
+    const link = screen.getByRole('img');
+    expect(link).not.toHaveAttribute('height', 14);
+  });
```
MkMan, line 25: I don't think this test's description and assertions match 🤔

src/components/GlobalNavigation/components/MobileView/components/NestedCollection/NestedCollection.spec.tsx
```diff
@@ -1,8 +1,113 @@
...
+
+  it('should render base component correctly', () => {
+    const { container } = render(
+      <TestBed items={NestedCollectionFixture.items} />,
+    );
+
+    expect(container).not.toBeEmptyDOMElement();
+  });
+
+  it('should click on nested collection', () => {
```
MkMan, line 83: This test description is a bit vague, why should it click on the nested collection?

src/components/GlobalNavigation/components/MobileView/components/NestedCollection/NestedCollection.spec.tsx
```diff
@@ -1,8 +1,113 @@
...
+
+    const nestedCollectionButton = screen.getByTitle('Nested Collection');
+
+    nestedCollectionButton.click();
+    const { onNestedCollectionClick } = useMobileViewContext();
+
+    expect(onNestedCollectionClick).toHaveBeenCalledTimes(1);
+    expect(listItemSpy).toHaveBeenCalledTimes(
+      NestedCollectionFixture.items.length + 1,
+    ); // TODO: Confirm why being called twice
```
MkMan, line 94: Any luck with this TODO?

src/components/GlobalNavigation/components/MobileView/components/Link/Link.spec.tsx
hassanraza0054: Really good catch. Found a data-testid was also here by default. So changed the test to use that, and its more robust now. Thanks @MkMan

src/components/GlobalNavigation/components/MobileView/components/Header/Header.tsx
hassanraza0054: Added "_BTN" suffixes ahead of them for a bit more context.

src/components/GlobalNavigation/components/MobileView/components/NestedCollection/NestedCollection.spec.tsx
hassanraza0054: Nice catch again. Added some more context.

src/components/GlobalNavigation/components/MobileView/components/NestedCollection/NestedCollection.spec.tsx
hassanraza0054: Yeah no luck atm so gonna keep it in as is for now. Hopefully when we come back to it, will have a look and see whats going on.

MkMan (approved):

PR #557, opened 2021-09-16
[Feature/CON-721] Updates to Google Map Slice

PixeldroidAU:
- ‘Theme’ field sets text to light when set to ‘Light’ and to dark when set to ‘Dark’
- Location field should dictate which field is used for Location data.


MkMan (approved):
This component is actually used 😮 ??

zack76:
> This component is actually used 😮 ??

i found it 
https://consideration.aesop.com/au/r/aesop-williamsburg/?contentful_env=matt-test-con&preview=true&contentful=true

zack76 (approved):

PixeldroidAU:
codeclimate is showing an issue however I don't think the recommended change is necessary. What does everyone think?

This and other components will be used in the new store pages.

PR #558, opened 2021-09-16
feat: remove name space for KR

zack76:
remove the space between first name and last name, especially for KR
Screen Shot 2021-09-16 at 17 06 26
Screen Shot 2021-09-16 at 17 06 28


MkMan:
Your branch has changes to Navigation and GoogleMaps 🤔 That doesn't seem right.

zack76:
> Your branch has changes to Navigation and GoogleMaps 🤔 That doesn't seem right.

Those changes actually come from `npm run format`. should be alright

MkMan:
> > Your branch has changes to Navigation and GoogleMaps 🤔 That doesn't seem right.
> 
> Those changes actually come from `npm run format`. should be alright

Oh! Someone has been cheeky and committed with `-n` 😅 Thanks @zack76

src/components/PersonalInfoSummary/PersonalInfoSummary.spec.tsx
```diff
@@ -45,10 +45,11 @@ describe('<PersonalInfoSummary />', () => {
     expect(screen.getByText(dateOfBirth)).toBeInTheDocument();
   });
 
-  it(`should reverse the order of the name if shouldSwapFullNameOrder is true`, () => {
+  it(`should reverse the order of the name and remove the name space if shouldSwapFullNameOrder and shouldRemoveNameSpace are true`, () => {
```
MkMan, line 48: Can you please test this in a new `it`?

src/components/PersonalInfoSummary/PersonalInfoSummary.types.ts
```diff
@@ -7,6 +7,7 @@ type PersonalInfoSummaryProps = {
   prefixOptions?: { value: string; text: string }[];
   shouldShowPrefix?: boolean;
   shouldSwapFullNameOrder?: boolean;
+  shouldRemoveNameSpace?: boolean;
```
MkMan, line 10: I remember Beau suggesting that using the affirmative was easier to understand. For example `shouldPutSpaceInName` and it can be defaulted to `true` if that's the common behaviour. I am not fussed and not sure which is best. What do you think?

zack76: Screen Shot 2021-09-17 at 10 22 31
actually this makes me confused. i'm not sure which is the better way too

src/components/PersonalInfoSummary/PersonalInfoSummary.spec.tsx
zack76: Screen Shot 2021-09-17 at 10 30 59
code climate throws this error when i use a new `it`.
any suggestion to avoid this kind of issue?

src/components/PersonalInfoSummary/PersonalInfoSummary.types.ts
zack76: i'd like to change it to `shouldHaveNameSpace = true`

hassanraza0054 (approved):
Looks good to me!

MkMan (approved):

src/components/PersonalInfoSummary/PersonalInfoSummary.spec.tsx
MkMan: You can click on the issues and mark them as Invalid or Wontfix
image

kevin-ho87 (approved):

PR #560, opened 2021-09-20
Add date selector

MkMan:
Extracted this [part](https://github.com/aesop/aesop-web-ui/blob/932a4e131df22a28262f4f1eab7d7ad2d0b76ba4/src/app/components-custom/account/information/AccountInfoPersonalForm.js#L145-L169) of Web UI into a component

DateSelector

### DIFF
--- a/src/components/DateSelector/DateSelector.tsx
+++ b/src/components/DateSelector/DateSelector.tsx
@@ -1,0 +1,89 @@
+import React, { useState, useCallback } from 'react';
+import cx from 'classnames';
+import { useThemeContext } from '~/contexts';
+import { Select } from '~/components/Select';
+import {
+  generateDayOptions,
+  generateMonthOptions,
+  generateYearOptions,
+  splitIsoDate,
+  useUpdateDayOptions,
+  useUpdateDayValue,
+} from './DateSelector.utils';
+import type { DatePortion, DateSelectorType } from './DateSelector.types';
+import styles from './DateSelector.module.css';
+
+const DateSelector: DateSelectorType = ({
+  copy,
+  maxYears = 100,
+  name = 'date-selector',
+  onChange,
+  theme,
+  value = '',
+}) => {
+  const currentTheme = useThemeContext(theme, 'dark');
+  const [date, setDate] = useState(value);
+  const { day, month, year } = splitIsoDate(date);
+  const [dayOptions, setDayOptions] = useState(generateDayOptions(31));
+
+  const classSet = cx(styles.base, styles[currentTheme]);
+
+  const updateDate = useCallback(
+    (portion: DatePortion, value: string) => {
+      const newDay = portion === 'day' ? value : day;
+      const newMonth = portion === 'month' ? value : month;
+      const newYear = portion === 'year' ? value : year;
+      const isNewDateValid = !!newDay && !!newMonth && !!newYear;
+
+      setDate(() => {
+        const newDate = [newYear, newMonth, newDay].join('-');
+        onChange?.(isNewDateValid ? newDate : '');
+        return newDate;
+      });
+    },
+    [day, month, onChange, year],
+  );
+
+  useUpdateDayOptions(date, setDayOptions);
+  useUpdateDayValue(date, updateDate);
+
+  return (
+    <div className={classSet}>
+      <Select
+        className={cx(styles.dropDown)}
+        label={copy?.day ?? 'Day'}
+        name={`${name}-day`}
+        onChange={(event) => {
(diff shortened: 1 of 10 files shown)


src/components/DateSelector/DateSelector.utils.ts
```diff
@@ -0,0 +1,87 @@
...
+  'April',
+  'May',
+  'June',
+  'July',
+  'August',
+  'September',
+  'October',
+  'November',
+  'December',
+];
```
kevin-ho87, line 34: Are there any localisation requirements?

src/components/DateSelector/DateSelector.tsx
```diff
@@ -0,0 +1,100 @@
...
+        className={cx(styles.dropDown)}
+        label={copy?.month ?? 'Month'}
+        name={`${name}-month`}
+        onChange={(event) => {
+          updateDate('month', event.target.value);
+        }}
+        options={generateMonthOptions(copy?.monthLabels)}
+        theme={currentTheme}
+        value={month}
+      />
```
kevin-ho87, line 84: Some countries have the month in first position. Is there a requirement for this?

src/components/DateSelector/DateSelector.tsx
```diff
@@ -0,0 +1,100 @@
...
+  maxYears = 100,
+  name = 'date-selector',
+  onChange,
+  theme,
+  value = '',
+}) => {
+  const currentTheme = useThemeContext(theme, 'dark');
+  const [date, setDate] = useState(value);
+  const { day, month, year } = splitIsoDate(date);
+  const [dayOptions, setDayOptions] = useState(generateDayOptions(31));
```
kevin-ho87, line 26: Would this need to accomodate the different months that have 30 and 28 days?

src/components/DateSelector/DateSelector.utils.ts
MkMan: There are, these are just used as fall-back (in the following line) in case the `copy` is not provided 🙂

src/components/DateSelector/DateSelector.tsx
MkMan: Definitely! This just sets an initial value of 31 days when the component mounts. I have a function further down to update the day options if the month/year changes to display the right number of days.

src/components/DateSelector/DateSelector.tsx
MkMan: I guess it's something we can look at implementing later if needed. I was modelling this heavily after the component I linked in the description which always had them as day-month-year.

src/components/DateSelector/DateSelector.utils.ts
```diff
@@ -0,0 +1,119 @@
...
+      value: i.toString(),
+      label: i.toString(),
+    });
+  }
+
+  return yearOptions;
+};
+
+const getMaxDaysInAMonth = (month: string, year: string): number =>
+  new Date(parseInt(year, 10), parseInt(month, 10), 0).getDate();
```
zack76, line 62: a quick question: why we need to use `parseInt`? I see it over several places. is it because we need to convert the string value to a number?

src/components/DateSelector/DateSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const classSet = cx(styles.base, styles[currentTheme]);
+
+  const updateDate = useCallback(
```
zack76, line 31: i still dont quite understand the use of `useCallback` here. could you explain a bit for me? thank you so much

src/components/DateSelector/DateSelector.utils.ts
MkMan: > is it because we need to convert the string value to a number?

Yes 👌

src/components/DateSelector/DateSelector.tsx
MkMan: Good point! It's actually no longer needed and can be removed now. I add it initially when the function was used inside a `useEffect`.

`useCallback` creates a function that only gets redefined when one of the variables specified in the dependency array changes (similar to how `useEffect` executes when values in its dependency array change). It's often used for performance optimisation. I used it initially however because eslint told me to 😬 

Screen Shot 2021-09-21 at 12 01 43 pm

zack76 (approved):

src/components/DateSelector/DateSelector.utils.ts
kevin-ho87: 👍

kevin-ho87 (approved):

PR #564, opened 2021-09-21
feat(Select): add isEnabled prop

MkMan:
Allow dropdowns to be disabled

Select


zack76 (approved):
👍🏼  surprised that we didn't have this before

PR #567, opened 2021-09-22
legacyMenu rename to isLegacyMenu

maxlew:
This PR renames the legacyMenu prop to isLegacyMenu

It also adds a `min-height: 0` rule to tablet view.

This legacyMenu hack/prop/feature allows us to mix the new and old navigation together by disabling or changing parts of the new Global Navigation while the old is active. Mostly this consists of disabling rendering any component and compressing the full screen take over. 

It's pure tech debt and should be removed once the final parts of the menu are migrated over.


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@aesop-engineering/gel",
+  "name": "aesop-gel",
```
MkMan: Won't this affect the package name published on npm registry?

MkMan (changes requested):
Thanks for renaming the variable 😌  🙏 

I wanted to verify the consequences of the [package name](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#name) change before we merge this.

package.json
MkMan: Never mind 😄

maxlew: already reverted, but that would have broken a lot of things :)

MkMan (approved):

hassanraza0054 (approved):

PR #568, opened 2021-09-22
Date selector fixes

MkMan:
Changes:
- Allow the component to be disabled by adding the `isEnabled` prop
- Make the component take 100% width of the element its in
- Fix an issue where onChange was called twice when the currently selected day value was higher than the number of days in the current month

### DIFF
--- a/src/components/DateSelector/DateSelector.tsx
+++ b/src/components/DateSelector/DateSelector.tsx
@@ -6,15 +6,16 @@ import {
   generateDayOptions,
   generateMonthOptions,
   generateYearOptions,
+  getMaxDaysInAMonth,
   splitIsoDate,
   useUpdateDayOptions,
-  useUpdateDayValue,
 } from './DateSelector.utils';
 import type { DatePortion, DateSelectorType } from './DateSelector.types';
 import styles from './DateSelector.module.css';
 
 const DateSelector: DateSelectorType = ({
   copy,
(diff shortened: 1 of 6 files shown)


src/components/DateSelector/DateSelector.tsx
```diff
@@ -45,12 +52,12 @@ const DateSelector: DateSelectorType = ({
   );
 
   useUpdateDayOptions(date, setDayOptions);
-  useUpdateDayValue(date, updateDate);
 
   return (
     <div className={classSet}>
       <Select
         className={cx(styles.dropDown)}
+        isEnabled={isEnabled}
```
zack76, line 60: just a bit curious in which scenario the selector will be disabled?

zack76 (approved):

src/components/DateSelector/DateSelector.tsx
MkMan: Very good question, I don't know 😅 I just figured since the Checkbox, Select and TextInput have it, this should too.

kevin-ho87 (approved):

src/components/DateSelector/DateSelector.tsx
zack76: ah, maybe these need to be disabled when the request is being processed after clicking the save button. sorry this is not related to this PR tho

PR #571, opened 2021-09-27
DynamicForm Enhancements

MkMan:
### Main changes:

- integrate DateSelector into DynamicForm
- allow`isEnabled` to be passed to form fields
- refactors and styling fix


kevin-ho87 (approved):

zack76 (approved):

src/compositions/DynamicForm/DynamicForm.types.ts
```diff
@@ -1,13 +1,18 @@
...
 import type { FieldValidation } from './validators/validators.types';
 
 type CheckBox = { type: 'Checkbox' };
-type Select = { type: 'Select'; options: SelectProps['options'] };
-type TextField = { type: 'TextField'; subtype: TextInputV2Props['type'] };
+type DateSelector = { type: 'DateSelector' } & Pick<
+  DateSelectorProps,
+  'copy' | 'maxYears'
+>;
+type Select = { type: 'Select' } & Pick<SelectProps, 'options'>;
```
hassanraza0054, line 14: The curly brace type setting im unfamiliar with. I know that **Pick** will have its type options inherited here. But as for the curly brace type, is 'Select' an actual default value type?

src/compositions/DynamicForm/FormField.tsx
```diff
@@ -36,29 +37,40 @@ export const FormField: VFC<FormFieldProps> = ({
...
+        ? fieldSpecificValues.maxYears
+        : undefined,
+    options:
+      fieldSpecificValues.type === 'Select'
+        ? fieldSpecificValues.options
+        : undefined,
+    subtype:
+      fieldSpecificValues.type === 'TextField'
+        ? fieldSpecificValues.subtype
+        : undefined,
```
hassanraza0054, line 70: Is it better if these undefined's would instead be nulls?

src/compositions/DynamicForm/wrappers/DateSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+  defaultValue,
+  isEnabled,
+  maxYears,
+  name,
+  rules,
+  theme,
+}) => (
+  <Controller
+    control={control}
+    defaultValue={defaultValue || ''}
```
hassanraza0054, line 23: If this is default date, might be better to change the name accordingly? Otherwise a bit confused what this is a value for. Keen to hear your thoughts.

src/compositions/DynamicForm/DynamicForm.types.ts
MkMan: I put together this TS [Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyhA2EDGwAKAnA9mAzlAvFAN4CwAUFFNsAJaYB2OAXFDsOjfQOYDaAugG5ylGgBMWbDtyFkAvuXKhIsBMmAFiSiCwDkcRCh2yoAMiioaSANYAefWozYcAGig7qdRjoB8M8kgY2VlUUFnsUDVIKKC1dcOAdZ2EqMFpAlh4dAEZEtwAmXJ0AZh0+clkgA) to demonstrate it.


> is 'Select' an actual default value type?

Sort of. It isn't a default value per se. It's a [string literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) which I am using to form a discriminated union type on the field `type`. I added a small demo to the previous Playground [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyhA2EDGwAKAnA9mAzlAvFAN4CwAUFFNsAJaYB2OAXFDsOjfQOYDaAugG5ylGgBMWbDtyFkAvuXKhIsBMmAFiSiCwDkcRCh2yoAMiioaSANYAefWozYcAGig7qdRjoB8MxeGgAYQALZCsAI0wADw0iLV0QsMiooz8yLSgAMRoEUQ1E62SoAB8VA2A0pAY2VlUUFmzc2OEoeLd7Q2cWj2qWHh0ARh1XHQAmYbcAZh0+cnkycirGdSRQwuiGnPg8wlIKVoCEtYjo4ZaAenOqMFpeqH6hkfGR6b4oS6gkAEMAVxwIPBfehQCDoLDoKA0ABmUC+olEEDy4WQv3+cyAA) to showcase the intent.

src/compositions/DynamicForm/FormField.tsx
MkMan: That would cause an error. These values get passed to components as props that are marked as optional (`?`). Optional in TS means either a certain type or undefined. e.g. `id?: string` will allow strings or an `undefined`. Passing `null` would violate that and throw a TS error. I put together this small [demo](https://www.typescriptlang.org/play?#code/C4TwDgpgBACgTgezAZygXigbwLACgpQCWAJgPwBcUywchAdgOYDceAvnngMYJ3VQBGAQ2LwkySqJToseAiUp0ArgBtlbDrm69gUBggQjEKCUdQYc+IsUqK6xCADN6EYmyA).

src/compositions/DynamicForm/wrappers/DateSelector.tsx
MkMan: These Wrapper components wrap underling form components in GEL and because their purpose is the same I am trying to keep them similar in their implementation. In the other wrappers ([TextField](https://github.com/aesop/aesop-gel/blob/develop/src/compositions/DynamicForm/wrappers/TextField.tsx#L26), [CheckBox](https://github.com/aesop/aesop-gel/blob/develop/src/compositions/DynamicForm/wrappers/Checkbox.tsx), [Select](https://github.com/aesop/aesop-gel/blob/develop/src/compositions/DynamicForm/wrappers/Select.tsx#L23)) it's called `defaultValue`.

If I call it `defaultDate` here, I am not sure what to call it in the other places 😅

PR #574, opened 2021-09-29
[Feature/CON-627]: Implement Sticky Nav

hassanraza0054:
Ticket: https://aesoponline.atlassian.net/browse/CON-627

## Summary
Objective was to replicate sticky nav functionality similar to what we already have on our native website. Key functionality include:-

1. Sticky nav activating when crossing the height of the nav + any content above (e.g. banners)
2. Sticky nav staying invisible as we keep scrolling down past the `navheight + offsetTop` threshold
3. The moment we scroll up while past this threshold, the sticky nav becomes visible
4. Nav resetting to default state with the help of the IntersectionObserver API if we scroll below the threshold

To recreate this behaviour, a few key choices were made:-

- A scroll listener event is created which starts running once the page is loaded, and changes state whenever we either **cross** a height threshold or **return** to one. 
- State being managed & referred to above is **stickyNavProps** which is centralised as a custom hook in the **GlobalNavigation.hooks.ts** file called `useStickyNav()`. This hook consumes a `stickyNavProps` object & `setStickyNavProps` function, returning a `stickyNavRef` and updating the sticky state accordingly. The state object has:
  - `isFixed`: Checks whether nav is fixed. **True** when crossing nav threshold
  - `isHidden`: Checks whether **invisible** or **visible** sticky nav based on whether we're scrolling **DOWN** or **UP** past the threshold
  - `offsetTop`: Calculates the height offset above wherever the nav is positioned. This is to capture the banner heights, whether **none**, **1** or **2** (is we dont have all banners for each market
- A `stickyNavRef` is created for each viewport, that attaches itself on an invisible div with absolute positioning with the **absoluteBuffer** class.
  - This ref is used for two purposes:- 
    -  Calculate the `offsetTop value` of the `parent div` wrapping all views. I've added this `relative`  positioned div container for all views in the `globalNavigation.tsx` component. Capturing this parent div, I use its `offsetTop` value as part of my threshold calculation formula for when the sticky nav should proc. This keeps consistency across all views, irrespective of the number of banners in view, as to how far one has to scroll before a nav becomes **Sticky**
    - Capture the absolute buffer div to manage the Sticky Nav state reset using the IntersectionObserverAPI
  - Finally, I added `isVisibleSticky` & `isInvisibleSticky` classes to each view, using some of these state variables as conditionals in order to achieve the sticky nav status.
   
That's a brief on the assumptions + changes I took on this PR. Feel free to ask me any more questions if you have any OR if i've missed anything! Happy to elaborate and discuss. 
 
## Video

https://user-images.githubusercontent.com/42990483/135324038-b1539dc6-9386-42b1-b4d5-015ca8c598d2.mov

## ⭐ 🕺  Conclusion [It sticks!] 🕺 ⭐ 
it sticks

## 🌟 ☝️ Now its your turn [Please review!] ☝️ 🌟
now its your turn


zack76 (commented):
awesome work!! just have a few questions

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
```diff
@@ -25,4 +26,74 @@ const useActiveView = (): { activeView: ActiveViewTypes } => {
...
+            isFixed: true,
+            isHidden: true,
+          });
+        }
+      }
+
+      prevScrollY.current = currentScrollY;
+    };
+
+    window.addEventListener('load', handleLoad);
```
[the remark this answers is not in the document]
hassanraza0054: Just added a description explaining its purpose. 

Basically what was happening when I refreshed a page was that, lets say I was already scrolled at `200px` down, which is past the threshold to apply sticky, the browser would do some auto-scrolling to set that position on refresh. I consistently saw that the last scroll would be past the threshold but **UPWARDS**, thus resulting in the sticky nav showing at the start of a refreshed page, which is not the same as our current behaviour. 

So the `isLoaded` serves the purpose of only applying that scroll listener logic once the page has loaded, so we can bypass the effects of auto-scrolling. Happy to show you this example in action if you're curious!

src/components/GlobalNavigation/components/DesktopView/DesktopView.tsx
```diff
@@ -53,6 +60,14 @@ const DesktopView: DesktopViewType = ({ className }) => {
   const classSet = cx(
     styles.base,
     { [closedClassName]: !isOpen },
+    {
+      [styles.isVisibleStickyNav]:
+        stickyNavProps.isFixed && !stickyNavProps.isHidden && !isOpen,
```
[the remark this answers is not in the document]
hassanraza0054: I'm not sure of a way to do this at the moment, as having both seperate seemed necessary to manage introduction the sticky nav. 

If I purely applied simply `isHidden` as I was originally trying to do, it applies the fixed styling right at the beginning of the scroll. 

I need another state, i.e. `isFixed` to identify once we've reached the **threshold** to apply the sticky nav. So a combination of these two felt like the best, most intuitive solution to achieve the desired result. Ofcourse, happy to take any feedback if you have changes in mind that achieve the same with one variable!

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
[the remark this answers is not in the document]
zack76: ah, didn't even notice this kind of behavior. so the auto-scrolling by the browser is not a scroll event, isn't it?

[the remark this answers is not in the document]
hassanraza0054: Yup. It actually fired the scroll event about 4 times without any interaction from me when refreshing. After adding the `isLoaded` value, I remove the effects those 4 fires could've had entirely. I say 4 as it was fixed on the page I was testing, but I imagine it might be a different number of scroll fires, maybe based on how elements render on the page.

src/components/GlobalNavigation/GlobalNavigation.types.ts
```diff
@@ -117,12 +117,21 @@ type GlobalNavigationStateContextProviderType = ComponentWithChildren<GlobalNavi
 
 type ActiveViewTypes = 'none' | 'mobile' | 'tablet' | 'desktop';
 
+type StickyNavType = {
+  isFixed: boolean;
+  isHidden: boolean;
+  topOffset: number;
+  isLoaded: boolean;
```
MkMan, line 124: In the description you mentioned this "Checks for when the page has fully loaded". The other props are referring to the nav bar so this could be mistaken to be referring to the nav bar having loaded. If that's the case, can it be renamed to something a more specific?

src/components/GlobalNavigation/components/DesktopView/DesktopView.module.css
```diff
@@ -14,7 +14,7 @@
   justify-content: space-between;
   padding-right: var(--container-margin);
   padding-left: var(--container-margin);
-  pointer-events: none;
+  pointer-events: auto;
```
MkMan, line 17: Will this change be fine when menus open? I am assuming this was there to prevent the user from interacting with other parts of the UI

src/components/GlobalNavigation/components/TabletView/TabletView.module.css
```diff
@@ -188,4 +207,4 @@
 
 .promotion {
   margin-bottom: 35px;
-}
+}
```
MkMan, line 210: Lost the new line here too 🤔

src/components/GlobalNavigation/components/DesktopView/DesktopView.tsx
```diff
@@ -17,25 +18,31 @@ import type { DesktopViewType } from './DesktopView.types';
...
+
   const {
     closedClassName,
     closedLogoTheme,
     openClassName,
   } = useDesktopViewContext();
 
+  const stickyNavRef = useRef();
+
+  useStickyNav(stickyNavRef, stickyNavProps, setStickyNavProps);
```
MkMan, line 44: I noticed you do this in all places you use the hook. Would it help (does it make sense) to have the hook create the ref and return it? That way it's done in 1 place and consumption is easier.

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
```diff
@@ -25,4 +26,72 @@ const useActiveView = (): { activeView: ActiveViewTypes } => {
...
+
+    if (stickyNavProps.isLoaded) {
+      window.addEventListener('scroll', handleScroll, { passive: true });
+    }
+
+    return () => {
+      window.removeEventListener('load', handleLoad);
+      window.removeEventListener('scroll', handleScroll);
```
maxlew, line 92: it might be worth debouncing this otherwise it's going to fire a lot.

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
```diff
@@ -25,4 +26,72 @@ const useActiveView = (): { activeView: ActiveViewTypes } => {
...
+  stickyNavProps: StickyNavType,
+  setStickyNavProps: (stickyNavProps: StickyNavType) => void,
+): void => {
+  const prevScrollY = useRef(0);
+
+  useEffect(() => {
+    const handleLoad = () => {
+      setStickyNavProps({ ...stickyNavProps, isLoaded: true });
+    };
+    const handleScroll = () => {
```
maxlew, line 40: Can this be defined outside of the useEffect?

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
```diff
@@ -25,4 +26,72 @@ const useActiveView = (): { activeView: ActiveViewTypes } => {
   return { activeView };
 };
 
-export { useActiveView };
+const useStickyNav = (
+  stickyNavRef: MutableRefObject<HTMLElement>,
+  stickyNavProps: StickyNavType,
+  setStickyNavProps: (stickyNavProps: StickyNavType) => void,
```
maxlew, line 32: Seeing as we're all in on context for this feature would it make more sense to include the `useGlobalNavigationStateContext` hook instead of passing props?

maxlew (commented):
Looks good probably worth having a catchup and run through of how it's all working before we merge it in.

src/components/GlobalNavigation/components/DesktopView/DesktopView.tsx
hassanraza0054: That would definitely make it leaner actually. Good catch! Changing now.

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
hassanraza0054: Keen to get some assistance on applying debounce if possible.

src/components/GlobalNavigation/components/DesktopView/DesktopView.module.css
hassanraza0054: So its interesting. What you're describing is actually achieved by `pointer-events: auto`. I also slipped up on this yesterday because the wording seems so counter-intuitve.

In the end, the best way I imagine it is to thing of the element as being **permeable** or **impermeable** with regards to pointer events. When `pointer-events: none`, the element becomes **permeable**, thus allowing the banners it is covering to be interactable. `pointer-events: auto` makes the element **impermeable**, so we therefore fix the issue by not allowing the banner links to be clickable through the element. Hope that clarifies this change.

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
MkMan: If it helps, [here](https://github.com/aesop/aesop-gel/blob/de05a279afce01754055a9264f31bf84655e400e/src/customHooks/useWindowHasResized/useWindowHasResized.ts#L13-L20) is an example usage

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
hassanraza0054: So I can't directly use the hook here as it would require applying the provider component as well. Instead, I've opted for abstracting the data out of mobile, tablet and desktop views and into the `GlobalNavigation.context.tsx` component.

src/components/GlobalNavigation/GlobalNavigation.types.ts
hassanraza0054: Removing the loaded logic as it was actually introducing problems with the listener.

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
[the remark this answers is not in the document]
hassanraza0054: Turns out I needed to remove this. Was causing issues. Having the scroll in as is works alright without the load listener 👍

src/components/GlobalNavigation/GlobalNavigation.module.css
```diff
@@ -0,0 +1,3 @@
+.base {
+  position: relative;
+}
```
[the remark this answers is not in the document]
hassanraza0054: Fixed!

src/components/GlobalNavigation/components/TabletView/TabletView.module.css
hassanraza0054: Fixed!

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
hassanraza0054: Debounce has some issues based on our current functionality. Wanted to close this PR today but had a crack at debounce first to see how it'd work with our nav. Results seemed concerning. We save on calls but there's potential cases for a janky nav experience.

Throttling might be better, but still introduces a somewhat janky experience. Tried our original nav and it has no throttling nor debounce. Not ideal, but thats how we get the smooth animation in and out after scrolling in each direction. Also, the nav only snaps responsively if we don't introduce debounce or throttle. 

So yeah, user experience takes a bit of a hit here. Tried all combinations and just feels like a bit of a design compromise. Keen to get thoughts next week on this @maxlew and @aesop/feds

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
hassanraza0054: Done!

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
hassanraza0054: Scratch what I said above. Instead opted to keep each view's refs distinct by getting it created through themselves with use of the sticky nav custom hook. Still passing props through to each view, but I feel its worth it as semantically, everything is where it's meant to be, making the code better documented.

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
hassanraza0054: Ignore what I said above. After a bit of digging, I found a much more optimised way to manage the sticky Nav using the IntersectionObserver API. Experience is all good now! Keen to show this next week.

src/components/GlobalNavigation/GlobalNavigation.context.tsx
```diff
@@ -21,12 +21,20 @@ const useGlobalNavigationStateStore: UseGlobalNavigationStateStore = ({
     initialActiveCollectionId,
   );
 
+  const [stickyNavProps, setStickyNavProps] = useState({
+    isFixed: false,
+    isHidden: false,
```
maxlew, line 26: Can it ever be both fixed and hidden at the same time? I guess from looking at the options there 4 possible states the menu can be in, it might make more sense to have something like 
```ts
state: 'fixed' | 'hidden' | 'default';
```

src/components/GlobalNavigation/components/DesktopView/DesktopView.tsx
```diff
@@ -17,27 +19,33 @@ import type { DesktopViewType } from './DesktopView.types';
...
   } = useGlobalNavigationContext();
 
+  const {
+    isOpen,
+    setIsOpen,
+    setActiveCollectionId,
+    stickyNavProps,
+    setStickyNavProps,
+  } = useGlobalNavigationStateContext();
+
```
maxlew, line 36: Is there a reason this moved below useGlobalNavigationContext? Just want to check there's no weird dependency to the call order of these.

src/components/GlobalNavigation/components/TabletView/TabletView.tsx
```diff
@@ -39,24 +43,41 @@ const TabletView: TabletViewType = ({ className }) => {
...
+        !stickyNavProps.isHidden &&
+        !isOpen,
+    },
+    {
+      [styles.isInvisibleStickyNav]:
+        !isCompletelyOnScreen &&
+        stickyNavProps.isFixed &&
+        stickyNavProps.isHidden &&
+        !isOpen,
+    },
```
maxlew, line 80: This code appears a lot, I guess it's a symptom of having three different components for each viewport. I wonder if we could abstract it away in future so it's returned from useStickyNav.

maxlew (approved):
Looks good, a couple of comments but nothing we need to change now.

src/components/GlobalNavigation/GlobalNavigation.utils.ts
```diff
@@ -3,8 +3,93 @@ import type {
   Link,
   NestedCollection,
   NotableNestedCollection,
+  StickyNavScrollType,
 } from './GlobalNavigation.types';
 
+/** Sticky Banner State Handlers
+ * 1. Sticky Nav is managed using the IntersectionObserver API, a buffer div & offset handlers
```
maxlew, line 10: Need to ensure this doesn't crash IE11, IntersectionObserver was never added to IE. If it just doesn't stick or is a bit degraded that's okay.

src/components/GlobalNavigation/components/DesktopView/DesktopView.tsx
hassanraza0054: Yeah so being created in the context introduces weird potential inconsistencies because, while changing viewports means unmounting that component, followed by mounting the new one (e.g. desktop to tablet view), in the old implementation, the ref was maintained as it was created in the globalNavigationContext. It didn't feel right, & I wasn't sure whether the refs would act weird, so I changed to add these props instead in each of the component views.

src/components/GlobalNavigation/GlobalNavigation.context.tsx
hassanraza0054: It can be! In the situation where you're scrolling passed the threshold but going down, its both **fixed** and **hidden**. 

It's important to set the class 'isInvisibleStickyNav' to facilitate the animation in. Whereas when its in the default state, all these sticky nav related classes get removed.

src/components/GlobalNavigation/components/DesktopView/DesktopView.module.css
```diff
@@ -57,6 +57,21 @@
...
+    height: 74px;
+    transition: top 300ms var(--easing-ease-in-out-cubic);
+  }
+
+  &.isVisibleStickyNav {
+    top: 0;
+  }
+
+  &.isInvisibleStickyNav {
+    top: -74px;
```
MkMan, line 72: There are some magic numbers here. Can you please put some comments explaining the logic for future devs.

src/components/GlobalNavigation/components/MobileView/MobileView.module.css
```diff
@@ -19,6 +19,28 @@
     opacity: 1;
   }
 
+  &.isVisibleStickyNav,
+  &.isInvisibleStickyNav {
+    position: fixed;
+    height: 90px;
```
MkMan, line 25: Similar with the `90px` here.

src/components/GlobalNavigation/components/TabletView/TabletView.module.css
```diff
@@ -36,6 +35,21 @@
...
+    height: 74px;
+    transition: top 300ms var(--easing-ease-in-out-cubic);
+  }
+
+  &.isVisibleStickyNav {
+    top: 0;
+  }
+
+  &.isInvisibleStickyNav {
+    top: -75px;
```
MkMan, line 50: How come this one is not `-74px` 🤔

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
```diff
@@ -1,7 +1,9 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useRef, MutableRefObject } from 'react';
 import { isViewport } from '~/utils/viewport';
 import { useWindowHasResized } from '~/customHooks';
-import type { ActiveViewTypes } from './GlobalNavigation.types';
+import type { ActiveViewTypes, StickyNavType } from './GlobalNavigation.types';
+import { stickyScrollHandler } from './GlobalNavigation.utils';
+import { debounce } from 'lodash';
```
MkMan, line 6: This will import all of lodash. Please import it as `import debounce from 'lodash/debounce';` Can you please also move this import up? We usually put 3rd party imports first 😃

src/components/GlobalNavigation/components/TabletView/TabletView.module.css
hassanraza0054: Yup theres a weird visual inconsistency where at exactly -74px, the nav's bottom part still slightly shows. So moved it an extra pixel up when invisible.

src/components/GlobalNavigation/components/MobileView/MobileView.module.css
hassanraza0054: Addressed in new changes! Thanks for the catch

src/components/GlobalNavigation/components/DesktopView/DesktopView.module.css
hassanraza0054: Yup made conversions for context! Thanks again.

src/components/GlobalNavigation/GlobalNavigation.hooks.ts
hassanraza0054: Really nice catch. Done!

MkMan (approved):

PR #593, opened 2021-10-18
[Breaking Change] Updates to bundling and installation

maxlew:
This is a fairly minor breaking change, but a breaking change none the less. 

There's two key things that have changed, the bundling and the building. Let's first go over the bundling.

## Bundling
Basically this project has always exported three formats 
- A single Common JS bundle
- A single ES Modules bundle
- A split Common JS bundle for each component 

This is why we see two different ways of importing things
top level import `import { Button } from '@aesop-engineering/gel'`
vs
direct import `import { Button } from '@aesop-engineering/gel/dist/components/Button`;

The first import is going to use one of the first two formats, normally with webpack or a modern build tool, it'll opt for the ES Modules. The second import is going to use the split bundle.

So the issue is that depending on the way you import gel, you might actually end up with two different versions entirely. 

This is further complicated because the single bundle exports a `main.css` file, while the per component bundle includes the css within each component. 

### So what changed?
Basically we removed the ES Modules bundle, and made the single Common JS bundle work the same as the split bundles removing so that it also contains the CSS. Under the hood, both the top level and direct imports actually reference the same built files, so no more two versions of gel.

## Building
This ones a bit easier, basically we've been trying to get rid of the dist file for a while but we need to it for linking 'in development' builds together. 

### So what's changed?
Now if you're installing a non npm version of this project, there's a 'prepare' step that will build aesop-gel during install. This will slow down the `npm i` command but it'll also mean no `dist` commited and all linking and feature branches are fine.


rollup.config.js
```diff
@@ -1,75 +1,117 @@
...
+const plugins = [
+  del({
+    targets: [
+      'dist',
+      // 'dist/assets',
+      // ...gelCategories.map((category) => `dist/${category}`),
+      // 'dist/sharedChunks',
+    ],
+    verbose: false,
+  }),
```
MkMan, line 67: Would we still need this plugin?

package.json
```diff
@@ -1,15 +1,14 @@
...
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
-  "style": "dist/styles.css",
+  "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
     "dist"
   ],
   "scripts": {
+    "prepare": "npm run type-definitions && rollup -c",
```
MkMan, line 11: This will run in GEL **before** publishing, is that correct? Is that why we're still keeping the `files` as `[ "dist" ]` ?

.releaserc
```diff
@@ -1,7 +1,7 @@
 {
-  branches: [
-    {name: 'main'},
-    {name: 'develop', prerelease: true},
+  "branches": [
+    { "name": "main" },
+    { "name": "next", "prerelease": true }
```
MkMan, line 4: This is why I usually prefer JS for config files over `.rc` 😄  Also comments 🙌

package.json
maxlew: it'll build aesop-gel during the `npm install` or `npm ci` of the project consuming it.

I feel like we can probably remove the files array, however let's just break one thing at a time for the moment.

PR #596, opened 2021-10-19
[Feature/CON-845 Add back button]

cyrusxzw:
-- Add 'isOpenSearchBackToMenu' in GlobalNavigation
-- Web-Ui will pass 'isOpenSearchBackToMenu' to 'MobileView.tsx' after clicking back button
-- 'isOpenSearchBackToMenu' is a boolean to control open Menu or not when travelling back from Search and Stores

### DIFF
--- a/src/components/GlobalNavigation/GlobalNavigation.types.ts
+++ b/src/components/GlobalNavigation/GlobalNavigation.types.ts
@@ -157,6 +157,7 @@ type GlobalNavigationContextType = {
   className?: string;
   collections: Collection[];
   isVisuallyObstructed?: boolean;
+  isOpenSearchBackToMenu?: boolean;
   /** User created on Navigation close event callback */
   onClose?: () => void;
   /** User created on Navigation open event callback */
@@ -190,6 +191,11 @@ type GetCollectionLists = (
   topLevelCollections: Link[];
 };
 
+type UseOpenMenuFromSearch = (
+  isOpenSearchBackToMenu: boolean,
+  setActiveCollectionId: (menu: string) => void,
+) => void;
+
 export type {
   Article,
   Actions,
@@ -214,4 +220,5 @@ export type {
   UseGlobalNavigationStateContext,
   UseGlobalNavigationStateStore,
   UseGlobalNavigationStore,
+  UseOpenMenuFromSearch,
 };
--- a/src/components/GlobalNavigation/components/MobileView/MobileView.tsx
+++ b/src/components/GlobalNavigation/components/MobileView/MobileView.tsx
@@ -1,7 +1,10 @@
 import React from 'react';
 import cx from 'classnames';
 import { ThemeContextProvider, useThemeContext } from '~/contexts';
-import { useStickyNav } from '../../GlobalNavigation.hooks';
+import {
+  useStickyNav,
+  useOpenMenuFromSearch,
+} from '../../GlobalNavigation.hooks';
 import {
   useEscapeKeyListener,
   useOnScreen,
@@ -32,6 +35,7 @@ const MobileView: MobileViewType = ({ className }) => {
     actions,
     collections,
     isVisuallyObstructed,
+    isOpenSearchBackToMenu,
     onClose,
     read,
     theme,
@@ -90,6 +94,8 @@ const MobileView: MobileViewType = ({ className }) => {
     actions.stores,
     actions.support,
   ];
+
+  useOpenMenuFromSearch(isOpenSearchBackToMenu, setActiveCollectionId);
 
   return (
     <ThemeContextProvider theme={currentTheme}>
--- a/src/components/GlobalNavigation/GlobalNavigation.hooks.ts
+++ b/src/components/GlobalNavigation/GlobalNavigation.hooks.ts
@@ -1,7 +1,7 @@
 import { useEffect, useState, useRef, MutableRefObject } from 'react';
 import { isViewport } from '~/utils/viewport';
 import { useWindowHasResized } from '~/customHooks';
-import type { ActiveViewTypes, StickyNavType } from './GlobalNavigation.types';
+import type { ActiveViewTypes, StickyNavType, UseOpenMenuFromSearch } from './GlobalNavigation.types';
 import { stickyScrollHandler } from './GlobalNavigation.utils';
 import throttle from 'lodash/throttle';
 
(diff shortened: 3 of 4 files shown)


src/components/GlobalNavigation/components/MobileView/MobileView.tsx
```diff
@@ -91,6 +92,10 @@ const MobileView: MobileViewType = ({ className }) => {
     actions.support,
   ];
 
+  useEffect(() => {
+    isOpenSearchBackToMenu && setActiveCollectionId('top');
+  }, [isOpenSearchBackToMenu]);
```
MkMan, line 97: There are a couple of [linting issues](https://github.com/aesop/aesop-gel/runs/3933884891?check_suite_focus=true#step:6:86) with these 2 lines. Should be easy to address 🙂

src/components/GlobalNavigation/GlobalNavigation.types.ts
```diff
@@ -157,6 +157,7 @@ type GlobalNavigationContextType = {
   className?: string;
   collections: Collection[];
   isVisuallyObstructed?: boolean;
+  isOpenSearchBackToMenu?: boolean;
```
MkMan, line 160: I don't understand what this variable does from its name. Can you please explain? 
Good boolean naming standards though 👍

src/components/GlobalNavigation/components/MobileView/MobileView.tsx
```diff
@@ -91,6 +92,12 @@ const MobileView: MobileViewType = ({ className }) => {
     actions.support,
   ];
 
+  useEffect(() => {
+    if (isOpenSearchBackToMenu) {
+      setActiveCollectionId('top');
```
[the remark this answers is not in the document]
cyrusxzw: I'll disable eslint. Cause every time isOpenSearchBackToMenu change, will result in dispatching setActiveCollectionId.

src/components/GlobalNavigation/components/MobileView/MobileView.tsx
```diff
@@ -91,6 +92,16 @@ const MobileView: MobileViewType = ({ className }) => {
     actions.support,
   ];
 
+  useEffect(
+    () => {
+      if (isOpenSearchBackToMenu) {
+        setActiveCollectionId('top');
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
MkMan, line 101: Nice! The linting warning is gone now 👍 
Could you please include a comment explaining why the rule was disabled? That way future developers know the intention.

```ts
    // eslint-disable-next-line react-hooks/exhaustive-deps -- explanation why the rule is disabled
```

hassanraza0054 (approved):

src/components/GlobalNavigation/components/MobileView/MobileView.tsx
```diff
@@ -91,6 +92,16 @@ const MobileView: MobileViewType = ({ className }) => {
     actions.support,
   ];
 
+  useEffect(
```
hassanraza0054, line 95: Could you actually quickly make this into a custom hook? That way, we stay consistent with the pattern of not just adding a blank useEffect into any of the views.

hassanraza0054 (changes requested):

src/components/GlobalNavigation/components/MobileView/MobileView.tsx
cyrusxzw: Good pick, I'll do it. ✌🏻

MkMan: What would the hook do? 🤔

cyrusxzw: Do the same as useEffect

src/customHooks/useOpenMenuFromSearch/useOpenMenuFromSearch.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from 'react';
+import type { UseOpenMenuFromSearch } from './useOpenMenuFromSearch.types';
+
+const useOpenMenuFromSearch: UseOpenMenuFromSearch= (isOpenSearchBackToMenu, setActiveCollectionId) => {
```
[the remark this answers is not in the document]
MkMan: I don't think that's correct 😅  There are existing [components](https://github.com/aesop/aesop-gel/search?l=TSX&q=useeffect) that use `useEffect` in their body.

If it's a standard in the Navigation to move `useEffect` out of components then I think this should remain within the Navigation folders. It's not reusable outside of the context of Navigation and so shouldn't be a global hook.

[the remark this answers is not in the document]
cyrusxzw: Yea, I think move it into Navigation folder is better. What do you think @hassanraza0054 ?

src/components/GlobalNavigation/components/MobileView/MobileView.tsx
hassanraza0054: I personally think it might be good to isolate @MkMan for two reasons:-

1. The hook could potentially be dealing with more variables in the future
2. From a visual consistency point, none of the views have a useEffect inside the nav components directly, but rather are using defined hooks (e.g. useFocusOnFirst) in a separate directory

src/customHooks/useOpenMenuFromSearch/useOpenMenuFromSearch.ts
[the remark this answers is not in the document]
hassanraza0054: Actually double minded about that now. If the team thinks we are better off with calling a direct useEffect, lets go with that @cyrusxzw . We can then add the useEffect directly in just mobile view.

src/components/GlobalNavigation/components/MobileView/MobileView.tsx
MkMan: In the picture you provided [here](https://github.com/aesop/aesop-gel/pull/596#discussion_r732294712) all these hooks are generic. Any component or even React app (Aesop or otherwise) can use them 🙌 

That's not the case with the functionality we're discussing. It's tightly coupled to Navigation and is not generic or reusable.

If the convention in Navigation is to extract `useEffect` code outside of components, that's ok, but the code should remain close to the component using it because it's coupled to it.

I hope that helps 😃

hassanraza0054 (approved):

PR #599, opened 2021-10-19
feat(NavigationBar): migrate to TypeScript

MkMan:


src/components/NavigationBar/NavigationBar.types.ts
```diff
@@ -0,0 +1,20 @@
+import type { ComponentWithoutChildren } from '~/types';
+
+type NavigationBarProps = {
+  childLinks?: {
+    hasTargetInNewWindow?: boolean;
+    text: string;
+    url: string;
```
zack76, line 7: do we have a specific type for the links

zack76 (approved):

src/components/NavigationBar/NavigationBar.types.ts
MkMan: Oooh good question, let me see if I can reuse the existing types.

PR #602, opened 2021-10-21
[Feature] Extra onChange handler for Dynamic Form

zack76:
1. extra onChange handler for Dynamic Form
2. allow dynamic form to have customised onChange handler

https://user-images.githubusercontent.com/51350376/138198994-9f70f594-4174-429a-b35e-f102b0c2a1b9.mov

### DIFF
--- a/src/compositions/DynamicForm/DynamicForm.types.ts
+++ b/src/compositions/DynamicForm/DynamicForm.types.ts
@@ -18,6 +18,8 @@ export type FieldSchema = {
   /** An initial value for the field. The value in `defaultValues` prop for the same field will take precedence */
   defaultValue?: string;
 
+  handleOnChange?: any;
+
   /** The field's HTML id attribute */
   id?: string;
 
--- a/src/compositions/DynamicForm/wrappers/Checkbox.tsx
+++ b/src/compositions/DynamicForm/wrappers/Checkbox.tsx
@@ -11,6 +11,7 @@ export const CheckboxWrapper: VFC<CheckboxWrapperProps> = ({
   dataTestRef,
   defaultValue,
   errorMessage,
+  handleOnChange,
   id,
   isEnabled,
   label,
@@ -31,7 +32,10 @@ export const CheckboxWrapper: VFC<CheckboxWrapperProps> = ({
         id={id}
         isEnabled={isEnabled}
         name={name}
-        onChange={(e) => onChange(e.target.checked)}
+        onChange={(e) => {
+          handleOnChange && handleOnChange(e);
+          onChange(e.target.checked);
+        }}
         theme={theme}
       />
     )}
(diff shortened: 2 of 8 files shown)


src/compositions/DynamicForm/wrappers/Checkbox.tsx
```diff
@@ -31,7 +32,10 @@ export const CheckboxWrapper: VFC<CheckboxWrapperProps> = ({
         id={id}
         isEnabled={isEnabled}
         name={name}
-        onChange={(e) => onChange(e.target.checked)}
+        onChange={(e) => {
+          handleOnChange && handleOnChange(e);
```
MkMan, line 36: Does this help?
```suggestion
          handleOnChange?.(e);
```

src/compositions/DynamicForm/DynamicForm.types.ts
```diff
@@ -18,6 +18,8 @@ export type FieldSchema = {
   /** An initial value for the field. The value in `defaultValues` prop for the same field will take precedence */
   defaultValue?: string;
 
+  handleOnChange?: any;
```
MkMan, line 21: `any` is not ideal, it would be best if we can specify a type. It might be easier if you do that on the individual types declared above as each is different.

What's the reason for choosing `handleOnChange` over just `onChange`? I thought we use `onChange` for props and `handleOnChange` when declaring on change functions.

src/compositions/DynamicForm/wrappers/Checkbox.tsx
zack76: oh damn yes

src/compositions/DynamicForm/DynamicForm.types.ts
zack76: i was thinking tracking or some clean up functions (such as removing error message) can be put in `handleOnChange` and it won't change any logic of `onChange' from the redux-hook-form `Controller`
Screen Shot 2021-10-21 at 13 40 55

MkMan: > it won't change any logic of onChange' from the redux-hook-form Controller

I didn't get what you mean, can you please explain?

zack76: Please check the below videos to see the differences

The error message won't get erased when the email address value is changed.
https://user-images.githubusercontent.com/51350376/138208536-fe651801-06ff-4870-892d-b92bd1791095.mov

In user creation flow, the error message will get erased after the email address value is changed.
https://user-images.githubusercontent.com/51350376/138208559-47d3d53b-4559-44fd-a247-e524bb1ed4e0.mov

MkMan: I remember facing this exact same issue 😭 Ended up putting an `onChange` listener on the form. It was disgusting and didn't work as well. Using onChange is definitely nicer 👌 

What I was asking about is that we usually call the prop `onChange` and the function passed to it either `onChange` or `handleOnChange`.

PR #627, opened 2021-11-08
[Fix] Update the style of modal

zack76:
1. Update the transition style of modal on mobile view
2. Update the style of close button
3. Update the style of checkbox error message

**Before**
Screen Shot 2021-11-08 at 11 39 16

**After**
Screen Shot 2021-11-08 at 11 38 59


src/components/Modal/Modal.tsx
```diff
@@ -46,7 +49,7 @@ const Modal: ModalType = ({
             isActive={isVisible}
             shouldMountOnEnter={true}
             shouldUnmountOnExit={true}
-            type="zoom"
+            type={isMobile ? 'slideDown' : 'zoom'}
```
hassanraza0054, line 52: Assuming you got a UI review on the new animation type. Nicely done 👌

hassanraza0054 (approved):

cyrusxzw (approved):

src/components/Checkbox/Checkbox.module.css
```diff
@@ -82,12 +82,7 @@
 
 .errorMessage {
   display: inline-block;
-  margin-left: 23px; /* box width + left margin */
   color: var(--color-error);
   font-family: var(--font-family-sans-serif-regular);
   font-size: rem(14px);
-
-  @media (--viewport-xs-sm-only) {
-    font-size: rem(12px);
```
MkMan, line 91: The error message will be larger than the actual checkbox text. Is that ok?

MkMan (approved):
All good as long as the UX team is good with these changes 😃

src/components/Checkbox/Checkbox.module.css
zack76: Ah, it wont happen. the size of both will be 14px

src/components/Modal/Modal.tsx
zack76: Yes, it's passed the UI review already and this change was requested by Michael actually.

src/components/Checkbox/Checkbox.module.css
MkMan: Oh, maybe I misunderstood the code on line 79 then 🤔

zack76: oh damn, i forgot i had updated the font size sin web ui. no need change here. Thank you!

kevin-ho87 (approved):

PR #636, opened 2021-11-16
feat(con-977): add store type eyebrow to store header

PixeldroidAU:
# Ticket Number

[CON-977](https://aesoponline.atlassian.net/browse/CON-977)

## Issue / Feature description

Update Store Detail Header to handle Eyebrow

"As a customer,
I would like a clear identifier to see what type of store page I am looking at,
To help me understand the store."

## Development tasks summary

- Store pages show a localize eyebrow field with the store type
- Store header styling tweak

## Screen grabs

Screen Shot 2021-11-16 at 3 07 30 pm


src/components/StoreDetailHeader/StoreDetailHeader.tsx
```diff
@@ -185,6 +186,16 @@ const StoreDetailHeader: StoreDetailHeaderType = ({
       className={classSet}
       content={
         <div className={wrapperClass}>
+          {storeType && (
+            <Heading
```
MkMan, line 190: Does this need to be a `h3`?

src/components/StoreDetailHeader/StoreDetailHeader.types.ts
```diff
@@ -21,7 +21,8 @@ type StoreDetailHeaderProps = {
   location?: string;
   openingHours?: HoursListItem[];
   phone?: string;
-  storeName: string;
+  storeName?: string;
+  storeType?: string;
```
MkMan, line 25: Thinking out loud but seeing that the value of this prop is used as an eyebrow and could be any string, would `eyebrow` be a suitable name for this prop?

src/components/StoreDetailHeader/StoreDetailHeader.types.ts
```diff
@@ -21,7 +21,8 @@ type StoreDetailHeaderProps = {
   location?: string;
   openingHours?: HoursListItem[];
   phone?: string;
-  storeName: string;
+  storeName?: string;
```
MkMan, line 24: Why did you make this prop optional?

src/components/StoreDetailHeader/StoreDetailHeader.tsx
PixeldroidAU: It doesn't _need_ to be h3, what would you suggest instead?

src/components/StoreDetailHeader/StoreDetailHeader.types.ts
PixeldroidAU: I think this is much clearer as to what the prop actually is. I think if we called it eyebrow that would be quite confusing. especially on the UI side of this.

src/components/StoreDetailHeader/StoreDetailHeader.types.ts
PixeldroidAU: I think I just did that out of habit. This prop should technically always be here but in the case it is not this will prevent an error.

maxlew (approved):
Looks good, happy to merge this in. There's a larger piece of work that's looking at headings and A11y especially around the eyebrow/headline pattern we use here.

src/components/StoreDetailHeader/StoreDetailHeader.tsx
MkMan: I think just a `span` will be fine 👍

src/components/StoreDetailHeader/StoreDetailHeader.types.ts
MkMan: If there isn't a requirement to change this then it's best to leave it as it was. An error being given by TS is an indication that something is wrong and that's helpful 😃

maxlew (approved):

PR #637, opened 2021-11-18
Feature/pur 610 rating]: Rating component

kevin-ho87:
# Ticket
https://aesoponline.atlassian.net/browse/PUR-610

### Description
Rating component

#### Reference
- https://www.w3.org/WAI/tutorials/forms/custom-controls/#a-star-rating
- https://mui.com/components/rating/

TODO: 
- theme

### Screenshot
image
with error

image


tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "paths": {
       "~/*": ["./src/*"]
     },
+    "downlevelIteration": true,
```
MkMan, line 9: What does this do? 🤔

src/components/Rating/Rating.types.ts
```diff
@@ -0,0 +1,17 @@
...
+
+type RatingProps = {
+  className?: string;
+  dataTestRef?: string;
+  errorMessage?: string;
+  theme?: Themes;
+  name: string;
+  value?: string;
+  onChange: (event: ChangeEvent<HTMLInputElement>) => void;
+  max?: number;
```
MkMan, line 12: Not a huge thing, but could you please sort these alphabetically? 😅 
Same with the de-structured props, please 🙏

src/components/Rating/Rating.types.ts
```diff
@@ -0,0 +1,17 @@
+import type { ChangeEvent } from 'react';
+import type { Themes, ComponentWithoutChildren } from '~/types';
+
+type RatingProps = {
+  className?: string;
+  dataTestRef?: string;
+  errorMessage?: string;
+  theme?: Themes;
+  name: string;
+  value?: string;
```
MkMan, line 10: Is this prop for a pre-selected value? For example to render the component with 3/5 starts selected?

src/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,56 @@
...
+  className,
+  theme = 'dark',
+  name,
+  value,
+  onChange,
+  dataTestRef,
+  errorMessage,
+  max = 5,
+}) => {
+  const currentTheme = useThemeContext(theme, 'dark');
```
MkMan, line 23: The default value is not needed, right?

```suggestion
  const currentTheme = useThemeContext(theme);
```

MkMan (commented):
This is **really** nice 👏 👏 👏 
I love how it's accessible and how you made the number of stars dynamic, and the tests 👌

src/components/Rating/Rating.types.ts
kevin-ho87: yes that is correct

src/components/Rating/Rating.tsx
kevin-ho87: yeah that is correct. it can be dark by default

tsconfig.json
kevin-ho87: TS was giving a warning when I used
```
[...Array(max).keys()]
```

src/components/Rating/Rating.types.ts
MkMan: Cool 👍 
I am thinking out loud, would `number` be an easier type to work with for developers consuming the component? It will be the same type as `max`

tsconfig.json
MkMan: This is the [documentation](https://www.staging-typescript.org/tsconfig#downlevelIteration) for it. I still don't get it 😂 but hey if it works 👍

MkMan (approved):

PR #644, opened 2021-11-29
[Feature/pur 610 textarea]: textareaV2 component

kevin-ho87:
TextareaV2 component.
Contains animating label, error message

image

image

image


src/components/TextareaV2/TextareaV2.fixture.ts
```diff
@@ -0,0 +1,3 @@
+const TextareaV2Fixture = {};
```
MkMan, line 1: Is this file needed if it's empty?

MkMan, line 1: Is this file needed if it's empty?

src/components/TextareaV2/TextareaV2.tsx
```diff
@@ -0,0 +1,89 @@
...
+      onChange,
+      rows,
+      theme = 'dark',
+      value: valueProp,
+    },
+    ref,
+  ) => {
+    const uid = useUID();
+    const textareaId = idProp || uid;
+    const [value, setValue] = useState(valueProp || '');
```
MkMan, line 28: Would using nullish coalescing be better? Otherwise a falsy value of `0` would end up being `''`

```suggestion
    const [value, setValue] = useState(valueProp ?? '');
```

MkMan (commented):
This is some fine work 👌 

Just one minor thing, when I ran it locally I couldn't notice a difference in styling when the component was enabled vs disabled

Textarea disabled

src/components/TextareaV2/TextareaV2.fixture.ts
kevin-ho87: no its not needed. i will delete

cyrusxzw:
Screen Shot 2021-11-30 at 9 46 27 am

I run at my end, and text is cropped a bit at the top in mobile view.

cyrusxzw:
Nice work 👍. Just wonder how to fix 'Similar blocks of code found in 2 locations' in codeclimate. I have the same issue in my PR.

kevin-ho87:
> Nice work 👍. Just wonder how to fix 'Similar blocks of code found in 2 locations' in codeclimate. I have the same issue in my PR.

Thanks, yeah its tricky. I am kinda testing different things

MkMan:
> Nice work 👍. Just wonder how to fix 'Similar blocks of code found in 2 locations' in codeclimate. I have the same issue in my PR.

I'd only worry about it if it's highlighting similar code blocks in the same file, otherwise don't stress 👍 

FYI @kevin-ho87 @cyrusxzw

MkMan (approved):

PR #645, opened 2021-11-29
Storybook improvements

MkMan:
### Changes

1. Update to the latest version of StoryBook
2. Migrate from the deprecated `-s` flag to serve static files. More [here](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated---static-dir-cli-flag)
3. Add a Favicon (because why not 😄)
4. Fix fonts path

#### Favicon
Screen Shot 2021-11-30 at 7 56 40 am

#### Fonts error before the change
Screen Shot 2021-11-30 at 7 52 36 am


maxlew (approved):
Looks good, I wonder if there's a way to make storybook a separate module inside of gel rather than having all it's dependencies inside of the main package.json... a problem for another day though.

PR #646, opened 2021-12-06
feature/con-1061

cyrusxzw:
#  Ticket: 
CON-1061 https://aesoponline.atlassian.net/browse/CON-1061

Fifty fifty floating text module MVP

##  Feature:

Have images interchange while scrolling text into view 

## Development tasks summary

Details can be found here: https://aesoponline.atlassian.net/browse/CON-999


src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,112 @@
...
+    textRefs.current = [];
+    const addToRefs = (element) => {
+      if (element && !textRefs.current.includes(element)) {
+        textRefs.current.push(element);
+      }
+    };
+    const handleScroll = throttle(() => {
+      const images = Array.from(
+        document.getElementsByClassName(
+          'fiftyScrollImage',
```
MkMan, line 34: Is this intentionally the same as the class assigned on line 72? If so, can you please put it in a constant so it doesn't break in the future if one is renamed.

MkMan (commented):
I reviewed some of it but I kept getting notified that the diff has changed. I will review the rest later.

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.types.ts
```diff
@@ -0,0 +1,19 @@
+import type { Themes } from '~/types';
+import type { ImageSectionProps } from './components/ImageSection/ImageSection.types';
+import type { TextSectionProps } from './components/TextSection/TextSection.types';
+
+type ActiveViewTypes = 'none' | 'mobile' | 'tabletanddesktop';
```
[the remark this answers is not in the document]
cyrusxzw: yep

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,112 @@
...
+      id,
+      percentageOfViewport = 0.4,
+      textBlocks = [],
+      theme = 'dark',
+    },
+    ref,
+  ) => {
+    const { activeView } = useActiveView();
+    const textRefs = useRef([]);
+    textRefs.current = [];
```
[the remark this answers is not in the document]
cyrusxzw: This is for initial. When change view from mobile to desktop, it will generate duplicated refs. So need initial again.

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,112 @@
...
+    useEffect(() => {
+      if (isInBrowser() && activeView === 'tabletanddesktop') {
+        window.addEventListener('scroll', handleScroll);
+      }
+      return function cleanup() {
+        if (isInBrowser() && activeView === 'tabletanddesktop') {
+          window.removeEventListener('scroll', handleScroll);
+        }
+      };
+    });
```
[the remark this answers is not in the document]
cyrusxzw: Yea, it will. Same here  https://github.com/aesop/aesop-gel/blob/9b2465798fffbc833dd1ced295b087ccae05d0ad/src/components/FullWidthHeroScroll/FullWidthHeroScroll.tsx#L34

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    const classImageSection = cx(styles.backgroundElement, 'fiftyScrollImage');
```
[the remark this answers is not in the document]
cyrusxzw: Good point!

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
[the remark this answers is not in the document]
MkMan: Will it still work if it only ran once (empty dependency array)? Are there issues with running `window.addEventListener` on every render cycle?

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { isInBrowser } from '~/utils/environment';
+import throttle from 'lodash/throttle';
+import { MobileView } from './MobileView';
+import { ImageSection } from './components/ImageSection';
+import { TextSection } from './components/TextSection';
+import styles from './FiftyFiftyFloatingText.module.css';
+import type { FiftyFiftyFloatingTextProps } from './FiftyFiftyFloatingText.types';
+import { useActiveView } from './FifityFiftyFloatingText.hooks';
+
+const FiftyFiftyFloatingText = forwardRef<any, FiftyFiftyFloatingTextProps>(
```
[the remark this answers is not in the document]
cyrusxzw: I change from section to div. And forgot to change the type.

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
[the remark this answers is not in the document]
cyrusxzw: It will not work if add []. Has to run every time to check scroll. I don't think there is an issue. Have researched it online and seems like it's a common solution. https://codesandbox.io/s/useeffect-scroll-event-oolh6?from-embed=&file=/src/index.js

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,119 @@
+import React, { useEffect, forwardRef, useRef } from 'react';
+import cx from 'classnames';
+import { isInBrowser } from '~/utils/environment';
+import throttle from 'lodash/throttle';
+import { MobileView } from './MobileView';
+import { ImageSection } from './components/ImageSection';
+import { TextSection } from './components/TextSection';
+import styles from './FiftyFiftyFloatingText.module.css';
+import type { FiftyFiftyFloatingTextProps } from './FiftyFiftyFloatingText.types';
+import { useActiveView } from './FifityFiftyFloatingText.hooks';
```
MkMan, line 10: There is a typo in the hooks file name 😄

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,119 @@
...
+    useEffect(() => {
+      if (isInBrowser() && activeView === 'tabletAndDesktop') {
+        window.addEventListener('scroll', handleScroll);
+      }
+      return function cleanup() {
+        if (isInBrowser() && activeView === 'tabletAndDesktop') {
+          window.removeEventListener('scroll', handleScroll);
+        }
+      };
+    });
```
MkMan, line 77: Seeing that there is already a separate file for hooks for this component, what do you think about moving this hook there too?

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const imageSectionClassSet = cx(styles.backgroundElement, scrollImageClass);
```
MkMan, line 79: I can't find the class `backgroundElement` in the css file 🤔

src/components/FiftyFiftyFloatingText/MobileView/MobileView.types.ts
```diff
@@ -0,0 +1,18 @@
...
+  backgroundImages?: {
+    image?: ImageSectionProps['image'];
+  }[];
+  textBlocks?: {
+    text?: TextSectionProps['text'];
+  }[];
+  theme?: Themes;
+};
+
+type MobileViewType = ComponentWithChildren<MobileViewProps>;
```
MkMan, line 16: I can't find the `children` prop used in this component. Should it instead be `ComponentWithoutChildren` ?

MkMan, line 16: I can't find the `children` prop used in this component. Should it instead be `ComponentWithoutChildren` ?

src/components/FiftyFiftyFloatingText/MobileView/MobileView.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div
+      className={cx(styles.base)}
+      data-testid="data-testid-FiftyFiftyFloatingTextMobile"
+    >
+      {backgroundImages.map(({ image }, index) => (
+        <div key={index}>
+          <ImageSection image={image} />
+          <TextSection
+            tabIndex={0}
```
MkMan, line 22: If `tabIndex` is always `0`, would it be better for the `TextSection` to hardcode it? 

I forgot, but what does `tabIndex` of 0 do? 😄

src/components/FiftyFiftyFloatingText/MobileView/MobileView.spec.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MobileView } from './MobileView';
+import { FiftyFiftyFloatingTextFixture } from '../FiftyFiftyFloatingText.fixture';
+
+describe('<MobileView />', () => {
+  it('should render mobile view correctly', () => {
```
MkMan, line 7: This test is a bit broad and doesn't test the functionality of the component. For example, if it rendered inputs instead of Images it would still pass.

One way to make it a bit more specific is to test it renders the correct number of `ImageSection`s and `TextSection`s. You can do this by spying/mocking those components and checking they were called.

src/components/FiftyFiftyFloatingText/components/ImageSection/ImageSection.tsx
```diff
@@ -0,0 +1,16 @@
...
+import styles from './ImageSection.module.css';
+import type { ImageSectionType } from './ImageSection.types';
+
+const ImageSection: ImageSectionType = ({ image, className }) => {
+  const classSet = cx(styles.base, className);
+  if (image) {
+    return <Image {...image} className={classSet} />;
+  }
+
+  return null;
```
MkMan, line 13: Nice! It's good to have a fallback if things are undefined 👌

src/components/FiftyFiftyFloatingText/components/ImageSection/ImageSection.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import cx from 'classnames';
+import { Image } from '~/components/Image';
+import styles from './ImageSection.module.css';
+import type { ImageSectionType } from './ImageSection.types';
+
+const ImageSection: ImageSectionType = ({ image, className }) => {
```
MkMan, line 7: This component doesn't have unit tests 😢

src/components/FiftyFiftyFloatingText/components/TextSection/TextSection.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { forwardRef } from 'react';
+import cx from 'classnames';
+import styles from './TextSection.module.css';
+import type { TextSectionProps } from './TextSection.types';
+
+const TextSection = forwardRef<HTMLDivElement, TextSectionProps>(
```
MkMan, line 6: This component doesn't have unit tests 😢

MkMan (commented):
Can you please include a video/gif of the component in the PR description. That way we know what it looks like 😃

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
cyrusxzw: oh my spelling 😭

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
cyrusxzw: yes

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
cyrusxzw: yea, you're right. I updated it and forgot to remove.

src/components/FiftyFiftyFloatingText/MobileView/MobileView.tsx
cyrusxzw: use for tab focus

src/components/FiftyFiftyFloatingText/components/ImageSection/ImageSection.tsx
cyrusxzw: The test of this component is covered in FiftyFiftyFloatingText.spec.tsx. I can see in in Gel, not every component has a unit test. But if it is a rule, I'll add.

src/components/FiftyFiftyFloatingText/components/TextSection/TextSection.tsx
cyrusxzw: Same with ImageSection

src/components/FiftyFiftyFloatingText/MobileView/MobileView.tsx
MkMan: `div`s are not intractable and so not focusable. It's possible to assign them a tabIndex of -1 to allow focus to be set only through JS.

From this MDN [page](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

> tabindex="0" means that the element should be focusable in sequential keyboard navigation, after any positive tabindex values and its order is defined by the document's source order

I guess that means that putting a tabIndex of 0 will make the `div` focusable. Is that intentional?

cyrusxzw: Yes, that's intentional. We need make the div focusable

src/components/FiftyFiftyFloatingText/components/ImageSection/ImageSection.tsx
MkMan: It's a not a rule although I would like it to be 😄

I personally think it's easier to test things individually than when used inside another unit of code.

Taking this component as an example, testing it alone will look roughly like:
1. test it renders nothing if `image` is not defined
2. test it renders the Image component if image is defined


Whereas testing this functionality when it's used inside another component would required satisfying all the conditions needed to make `ImageSection` run.
There is also the issue that the implementation is being tested further away from where it lives.

It's not a rule so you can decide what you wish to do 🙂

src/components/FiftyFiftyFloatingText/MobileView/MobileView.tsx
MkMan: Oh ok, why does it need to be focusable? 🤔

cyrusxzw: UX/Ui require

src/components/FiftyFiftyFloatingText/components/ImageSection/ImageSection.tsx
cyrusxzw: I'll add some tests.

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
cyrusxzw: I just found it may not be moved. It doesn't have a return value for a custom hook.

MkMan: That should be fine still. It can look something like

```ts
const useSetScrollHandler = (
  activeView: ActiveViewTypes,
  handleScroll: () => void,
) => {
  useEffect(() => {
    if (isInBrowser() && activeView === 'tabletAndDesktop') {
      window.addEventListener('scroll', handleScroll);
    }
    return function cleanup() {
      if (isInBrowser() && activeView === 'tabletAndDesktop') {
        window.removeEventListener('scroll', handleScroll);
      }
    };
  });
};
```

And used as 

```ts
useSetScrollHandler(activeView, handleScroll);
```

src/components/FiftyFiftyFloatingText/MobileView/MobileView.tsx
kevin-ho87: is this because the want the floating sections to be focusable?

src/components/FiftyFiftyFloatingText/components/ImageSection/ImageSection.module.css
```diff
@@ -0,0 +1,26 @@
+@import 'styles/index';
+
+.base.base {
+  position: absolute;
```
kevin-ho87, line 4: would there need to be position values like `top` `left` ? 
or is that set elsewhere

src/components/FiftyFiftyFloatingText/MobileView/MobileView.tsx
cyrusxzw: yes

src/components/FiftyFiftyFloatingText/components/TextSection/TextSection.module.css
```diff
@@ -0,0 +1,44 @@
...
+.base {
+  display: flex;
+  height: 100vh;
+  box-sizing: border-box;
+  align-items: center;
+  padding-left: 40px;
+  margin-right: 80px;
+  transform: translate3d(0, 0, 0);
+
+  & p {
```
kevin-ho87, line 12: you could do something like
```
> * { 
  margin......
}
```
just incase something non `p` is passed in

src/components/FiftyFiftyFloatingText/components/ImageSection/ImageSection.module.css
cyrusxzw: they are inherited from parent element

src/components/FiftyFiftyFloatingText/components/TextSection/TextSection.module.css
cyrusxzw: Assumption is p only. but good to change it to * for future. 👍

src/components/FiftyFiftyFloatingText/FiftyFiftyFloatingText.tsx
cyrusxzw: I did exact same code yesterday, and thought no return value, deleted it. Will add them back.

hassanraza0054 (approved):

PR #649, opened 2022-01-12
fix(accessibility): add screen reader text to all loading components

JasonPingAesop:
Ticket: https://aesoponline.atlassian.net/browse/PUR-559

Adding screen reader text to all loading indicator in order to meet accessibility requirement.


src/constants/accessibility.ts
```diff
@@ -0,0 +1 @@
+export const SCREEN_READER_TEXT = 'Loading, please wait.';
```
kevin-ho87, line 1: Would this mean that the loading text will always be 'Loading, please wait.' on all sites.
Is there any localisation requirements?

MkMan (commented):
Some of the components take a `copy` prop that contains the content they display. This is to facilitate different languages.

Where possible, can you please add a field for loading inside `copy` perhaps with a fallback value in English instead of hardcoding?

src/constants/accessibility.ts
hassanraza0054: Same question!

hassanraza0054 (commented):
Just got the same question as kevin regarding translations. Curious if this requires a contentful and GQL piece of work to get those in.

hassanraza0054:
> Some of the components take a `copy` prop that contains the content they display. This is to facilitate different languages.
> 
> Where possible, can you please add a field for loading inside `copy` perhaps with a fallback value in English instead of hardcoding?

Good catch here. Perhaps the constant declared atm can be that english fallback!

src/components/Audio/__snapshots__/Audio.spec.tsx.snap
```diff
@@ -28,6 +28,11 @@ exports[`<Audio /> should render base component correctly 1`] = `
         data-testid="data-testid-loading"
         role="status"
       >
+        <span
+          class="srOnly"
```
cyrusxzw, line 32: I suggest use 'screenReaderOnly', which is easier to read than abbreviation.

MkMan: I also don't like abbreviations but I think this is caused by existing code in the Loading component https://github.com/aesop/aesop-gel/blob/main/src/components/Loading/Loading.module.css#L49

src/constants/accessibility.ts
JasonPingAesop: Good catch up guys, I have removed the hard coded pieces. Allowing parent components can accept screen reader text to pass to loading components

src/components/Audio/__snapshots__/Audio.spec.tsx.snap
JasonPingAesop: True, they are css related

JasonPingAesop:
> Just got the same question as kevin regarding translations. Curious if this requires a contentful and GQL piece of work to get those in.

True, even in web-ui repo, most of screen reader texts are hard coded in English.

JasonPingAesop:
> Some of the components take a `copy` prop that contains the content they display. This is to facilitate different languages.
> 
> Where possible, can you please add a field for loading inside `copy` perhaps with a fallback value in English instead of hardcoding?

Maybe not fallback in english, as the loading components use `screenReaderText` variable to conditional render this text. Otherwise it will render the text forever.

src/components/AddToCartButton/AddToCartButton.types.ts
```diff
@@ -14,6 +14,7 @@ type AddToCartButtonProps = {
   dataTestRef: string;
   isEnabled?: boolean;
   isFullWidth?: boolean;
+  screenReaderText?: string;
```
MkMan, line 17: Being content presented to the user I thought `copy` would be a good place for this value. What do you think?

JasonPingAesop: Agree, I have moved it into copy variable.

MkMan (approved):

PR #654, opened 2022-01-14
fix Product Commerce component not displaying issue

cyrusxzw:
Ticket:
https://aesoponline.atlassian.net/browse/CON-622

It is a quick fix which will has minimum impact on the components using 'useOnScreen' hook. Ideal solution is to rebuild  'useOnScreen' hook by using callback ref. But if so, will need to pass function rather than ref that will be a huge work. 

ps: QA will test all the related components to make sure no breaks.


src/customHooks/useOnScreen/useOnScreen.ts
```diff
@@ -20,39 +20,40 @@ const useOnScreen: UseOnScreen = (
...
-        },
-      );
-
-      observer.observe(currentRef);
-    }
-
-    return function cleanup() {
-      if (hasIntersectionObserver) {
-        observer.unobserve(currentRef);
+    setTimeout(() => {
```
MkMan, line 23: What does wrapping it in `setTimeout` do?

src/customHooks/useOnScreen/useOnScreen.ts
```diff
@@ -20,39 +20,40 @@ const useOnScreen: UseOnScreen = (
...
+          {
+            rootMargin,
+            threshold,
+          },
+        );
+
+        observer.observe(currentRef);
       }
-    };
+      return function cleanup() {
```
MkMan, line 51: Can you please keep the line break above the cleanup function.

cyrusxzw: yep

src/customHooks/useOnScreen/useOnScreen.ts
cyrusxzw: To get ref after page content load.

MkMan (approved):

PR #656, opened 2022-01-17
build: remove css comments from individual component modules

MkMan:


hassanraza0054 (approved):
This only removes comments during compilation im assuming

MkMan:
> This only removes comments during compilation im assuming

Yes ✅

PR #652, opened 2022-01-13
chore: update commit tools to latest versions

MkMan:

PR #659, opened 2022-01-18
chore: exclude commit messages in CI from linting

MkMan:
Not sure why but the CI build started [failing](https://github.com/aesop/aesop-gel/runs/4860166580?check_suite_focus=true#step:10:222) recently when trying to commit because the commit message body had long lines. It might be due to the version increase done in #652.

This change should exclude commit messages made in the CI from commit linting.


commitlint.config.js
```diff
@@ -4,4 +4,5 @@ module.exports = {
     'header-max-length': [2, 'always', 101],
     'scope-case': [0],
   },
+  ignores: [(message) => message.includes('[skip ci]')],
```
MkMan, line 7: The commit messages made in the CI follow the template defined here https://github.com/aesop/aesop-gel/blob/main/.releaserc#L16

cyrusxzw (approved):

commitlint.config.js
MkMan: Configuration reference [here](https://commitlint.js.org/#/reference-configuration?id=javascript)

JasonPingAesop (approved):

maxlew (approved):
👍

commitlint.config.js
JasonPingAesop: Commitlint added new rules in config-conventional.js
[here](https://github.com/conventional-changelog/commitlint/blob/5a5f6419b6e89a32ef3faa80dc229e08a1e218af/%40commitlint/config-conventional/index.js#L5)
That's why it has limitation on body text message. We can set `body-max-line-length` in our commitlint.config.js

PR #660, opened 2022-01-19
[Feature/pur 554 modal]: refocus on element after modal close

kevin-ho87:
### Description
- install focusTrap [plugin](https://github.com/focus-trap/focus-trap-react)
- plugin enables focus to return to original element after closing
- allow overlay outside of focusTrap to be clicked


package.json
```diff
@@ -62,6 +62,7 @@
   "dependencies": {
     "@googlemaps/markerclustererplus": "^1.2.0",
     "classnames": "^2.2.5",
+    "focus-trap-react": "^8.9.1",
```
MkMan, line 65: Do we include this in the `peerDependencies` like the other `dependencies` in this project?

src/components/Modal/Modal.tsx
```diff
@@ -51,26 +43,31 @@ const Modal: ModalType = ({
...
-              aria-hidden={!isVisible}
-              aria-label={aria?.label}
-              aria-modal="true"
-              className={classSet}
-              data-testid="data-testid-Modal"
-              ref={ref}
-              role="dialog"
+            <FocusTrap
+              focusTrapOptions={{
+                allowOutsideClick: true,
```
MkMan, line 48: Would clicking outside close the modal?

src/components/Modal/Modal.tsx
```diff
@@ -2,20 +2,15 @@ import React from 'react';
...
-import { mergeRefs } from '~/utils/mergeRefs';
+import { useEscapeKeyListener, useOverflowHidden } from '~/customHooks';
 import { getPortalRoot } from '~/utils/portal';
 import { ModalBody } from './components/ModalBody';
 import { Overlay } from '~/components/Overlay';
 import { Transition } from '~/components/Transition';
 import { isViewport } from '~/utils/viewport';
 import type { ModalType } from './Modal.types';
 import styles from './Modal.module.css';
+import FocusTrap from 'focus-trap-react';
```
MkMan, line 13: Small thing but can you please move this import with the other 3rd party imports at the top of the file?

src/components/Modal/Modal.tsx
```diff
@@ -2,20 +2,15 @@ import React from 'react';
 import { createPortal } from 'react-dom';
 import cx from 'classnames';
 import { useThemeContext } from '~/contexts';
-import {
-  useEscapeKeyListener,
-  useFocusOnFirst,
-  useOverflowHidden,
-  useTrapFocus,
-} from '~/customHooks';
-import { mergeRefs } from '~/utils/mergeRefs';
```
MkMan, line 11: Do you know if the removed imports are used elsewhere in this repo or web-ui?

src/components/Modal/Modal.tsx
kevin-ho87: Yes it would, `allowOutsideClick` allows the modal background overlay to be clicked('close' function attached to overlay)

src/components/Modal/Modal.tsx
kevin-ho87: Hooks from `customHooks` are still used. 
`mergeRefs` does not seem to be used elsewhere.

MkMan: Hmmm 🤔 Do you think we should remove it?

package.json
MkMan: I'd say include it for consistency with the existing code.

src/components/Modal/Modal.tsx
kevin-ho87: Decided to leave it for now.

MkMan (approved):

PR #664, opened 2022-01-20
fix(con-570): fix zindex issue related to zendesk

masonwongcs:
## Summary
- Fix z-index issue related to [CON-570](https://aesoponline.atlassian.net/browse/CON-570) where the Zendesk button was layered on top of navigation in Mobile View.
- This fix added a new z-index variable for mobile and fix the layer issue


src/components/GlobalNavigation/components/MobileView/MobileView.module.css
```diff
@@ -13,7 +13,7 @@
 
 .base {
   position: relative;
-  z-index: var(--z-index-menu);
+  z-index: var(--z-index-mobile-menu);
```
cyrusxzw, line 16: Do you still need to add a new value '--z-index-mobile-menu' ? Cause --z-index-menu and --z-index-mobile-menu are both 100.

masonwongcs: I think it is better to make two copies of each for maintainability, it's easier to maintain next time if we separate it into two variables. 😄

MkMan (approved):

cyrusxzw (approved):

PR #680, opened 2022-02-17
feat(pur-764): add to cart disabled button

kevin-ho87:
`<AddToCartButton />` normally renders as hyperlink when `alternateAction` object is passed in.
This update renders a button instead when `alternateAction.isDisabled` is passed in.

image


MkMan (commented):
If where this component is used all of `url`, `label` and `isDisabled` are truthy, would you then end up with a disabled button and a link?

cyrusxzw (approved):

kevin-ho87:
> If where this component is used all of `url`, `label` and `isDisabled` are truthy, would you then end up with a disabled button and a link?

No, you would only get the disabled button

PR #687, opened 2022-02-24
Fix/con 1169

masonwongcs:
### To fix the sluggish animation in safari

Ticket: https://aesoponline.atlassian.net/browse/CON-1169


kevin-ho87:
Does this change affect the other browsers?

masonwongcs:
Yup, current transition of `height` will be replaced with `clip-path`.

cyrusxzw:
Good, I guess it's a nice transition.

cyrusxzw (approved):

PR #688, opened 2022-02-24
feat(react-hook-form): update to v7.27.1

MkMan:


src/compositions/DynamicForm/wrappers/DateSelector.tsx
```diff
@@ -25,7 +25,7 @@ export const DateSelectorWrapper: VFC<DateSelectorWrapperProps> = ({
     control={control}
     defaultValue={defaultValue || ''}
     name={name}
-    render={({ onChange: reactHookFormOnChange, value }) => (
+    render={({ field: { onChange: reactHookFormOnChange, value } }) => (
```
kevin-ho87, line 28: what is the difference with the extra `field` ?

kevin-ho87 (approved):

cyrusxzw (approved):

src/compositions/DynamicForm/wrappers/DateSelector.tsx
MkMan: The [API changed in version 7](https://react-hook-form.com/migrate-v6-to-v7), the `onchange` and field `value` are now nested inside a `field` object.

PR #689, opened 2022-02-28
Enhance checkbox

MkMan:
## Changes:

- add isChecked prop to allow setting a default value
- add disabled styling for UX
- move checkmark up to line up with box

## Before

image

## After
image


cyrusxzw:
What does this `disabled checkbox` use for?

MkMan:
> What does this `disabled checkbox` use for?

There is a use case in Web UI where onChange of the checkbox triggers an api call. To calling the API while the current call is pending we disable the checkbox.

cyrusxzw (approved):

src/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -33,4 +34,41 @@ describe('<Checkbox />', () => {
...
+      const onChange = jest.fn();
+
+      render(<Checkbox content="Check me" onChange={onChange} />);
+      const checkBox = screen.getByRole('checkbox');
+      userEvent.click(checkBox);
+
+      expect(onChange).toHaveBeenCalled();
+    });
+
+    it(`should not fail when an onChange callback is not povided`, () => {
```
kevin-ho87, line 49: Typo at `povided`

src/components/Checkbox/Checkbox.module.css
```diff
@@ -37,9 +37,25 @@
...
+      color: var(--color-dark-disabled);
+    }
+
+    &.light {
+      border-color: var(--color-light-disabled);
+      color: var(--color-light-disabled);
+    }
+
+    &:hover {
+      cursor: not-allowed;
```
kevin-ho87, line 52: You may want to check this hover state with a designer

MkMan: Yeah good point. I thought it would be better than still having cursor on a disabled input. I need these changes for a ticket I am working on so I'll revert this change and leave it until someone from UX brings it up.

src/components/Checkbox/Checkbox.spec.tsx
MkMan: Great pick up, thank you. It's fixed now.

kevin-ho87 (approved):

PR #691, opened 2022-03-07
Add description to checkbox

MkMan:
## Changes

- Add a description prop so consuming app doesn't have to line up the description manually (requires knowing how much to indent by)
- only add arai-describedby if there is an error (Axe shows an error if the attribute is there without an error)

image


src/components/Checkbox/Checkbox.tsx
```diff
@@ -46,7 +47,7 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
       <div>
         <label className={baseClassSet} htmlFor={inputId}>
           <input
-            aria-describedby={errorMessageId}
+            aria-describedby={errorMessage ? errorMessageId : undefined}
```
cyrusxzw, line 50: I think `aria-describedby` must point to an element. Can we do `undefined` on it?

MkMan: > I think aria-describedby must point to an element.

That's correct which is why Axe was complaining. 

> Can we do undefined on it?

We can set DOM attributes to `undefined`, React doesn't add them to the DOM in that case.

cyrusxzw (approved):

PR #692, opened 2022-03-07
feat(con-1226): Enhance Modal to support full width screen and customized background color

cyrusxzw:
# [CON-1226](https://aesoponline.atlassian.net/browse/CON-1226)

## Issue / Feature description
Modal cannot support full screen, and background color is set to light. Only close icon can be changed based off theme color.

## Development tasks summary
- Add an prop `isFullScreen` to trigger expand to full screen. 
- Add background color based off theme color.
- Change default theme color to `light`. So the default modal background color is light and close icon color is dark, which make more sense. 

## Screenshot

https://user-images.githubusercontent.com/29698458/157137171-1dfd5402-6212-4115-a2b4-3c0f11f45c42.mov


src/components/Modal/Modal.tsx
```diff
@@ -21,16 +21,20 @@ const Modal: ModalType = ({
   children,
   className,
   copy,
+  isFullScreen = false,
   isVisible,
   onClose,
   theme,
 }) => {
-  const currentTheme = useThemeContext(theme, 'dark');
+  const currentTheme = useThemeContext(theme, 'light');
```
MkMan, line 29: Why is the default theme changing?

MkMan (commented):
I noticed in the video that when the theme is `dark` the text is dark against a dark background.

src/components/Modal/Modal.tsx
cyrusxzw: The previous default theme is `dark`, which means close icon is in dark color. But now, I added background color. `dark` means background color is dark, and close icon is `light`. In order to keep background color is `light` and close icon is `dark` for the existing component using Modal, I change the default color to `light`.

src/components/Modal/Modal.types.ts
```diff
@@ -6,7 +6,9 @@ type ModalAndModalBodyProps = {
   aria?: {
     label?: string;
   };
+  backgroundColor?: string;
```
kevin-ho87, line 9: is this prop type used at all?

src/components/Modal/Modal.tsx
kevin-ho87: just make sure the pages currently using the modal still works as intended

src/components/Modal/Modal.types.ts
cyrusxzw: Oh, nope, I forgot to remove it. thanks

cyrusxzw:
> 

It doesn't matter, can always control the color of modal content.

src/components/Modal/Modal.tsx
cyrusxzw: Yea, I found `SavedCreditCardDeleteModal` and `FormModal` used Modal from Gel. They work well.

MkMan (approved):

kevin-ho87 (approved):

hassanraza0054 (approved):

PR #693, opened 2022-03-08
Rename Checkbox prop

MkMan:
## Changes
- rename content to label to be consistent with other form components in the library
- move the input from outside the label as per [a11y best practices](https://css-tricks.com/html-inputs-and-labels-a-love-story/#aa-how-to-pair-a-label-and-an-input)
- remove browser specific styling, our tooling should be adding them automatically
- remove top margin from the component. Shared components should not ship with spacing around them, instead consuming code should manage that

This is a **BREAKING CHANGE** and I will create sister PRs in projects that use GEL to update the code there.


src/components/Checkbox/Checkbox.tsx
```diff
@@ -35,7 +35,7 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
     };
 
     const baseClassSet = cx(styles.base, className);
-    const contentClassSet = cx(styles.content, styles[theme], contentClassName);
+    const contentClassSet = cx(styles.content, styles[theme], labelClassName);
```
kevin-ho87, line 38: Should this be called `labelClassSet` ?

MkMan: Yes, yes it should 😅 Great pick up, will update it now

kevin-ho87 (approved):

src/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -58,14 +58,14 @@ describe('<Checkbox />', () => {
...
-      render(<Checkbox content="Check me" isChecked={true} />);
+      render(<Checkbox isChecked={true} label="Check me" />);
       const checkBox = screen.getByRole('checkbox');
 
       expect(checkBox).toBeChecked();
     });
 
     it(`should render unchecked when isChecked is false`, () => {
-      render(<Checkbox content="Check me" isChecked={false} />);
+      render(<Checkbox isChecked={false} label="Check me" />);
```
hassanraza0054, line 68: Would it be worth to make the label into a reusable variable? Just a small quality of life change. Won't matter much since its a test file, but is probably slightly better than hardcoded text?

hassanraza0054 (approved):

src/components/Checkbox/Checkbox.spec.tsx
MkMan: Good point, it's a mandatory field so every test needs it. I'll give it a crack

MkMan: Done ✅

PR #698, opened 2022-03-23
Feature/ret 590 password input component

ivandaho:
https://aesoponline.atlassian.net/browse/RET-590

Password component which is built on top of `TextInputV2`.

It supports delayed password masking. The latest character typed will be shown, but previous characters will be masked. After a timeout, the latest character typed will be masked as well.

Add new props to `TextInputV2` to support `PasswordInput`
tests for password component as user is typing, currently `@testing-library/user-event` `userEvent.type()` does not trigger `onBeforeInput()` which is needed for delayed password masking to work. (see https://github.com/testing-library/user-event/issues/858)


src/components/PasswordInput/PasswordInput.module.css
```diff
@@ -0,0 +1,30 @@
...
+  }
+}
+
+.button {
+  @mixin inputFontSize;
+
+  position: absolute;
+  top: 25px;
+  right: 0;
+  line-height: 170%;
```
masonwongcs, line 17: I think you may use `line-height: 1.7` instead.

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
```diff
@@ -0,0 +1,130 @@
+import cloneDeep from 'lodash/cloneDeep';
+
+import { CursorRef } from './useDelayedPasswordMasking.types';
+
+/** replace `event.target.value` with custom value */
+export const makeChangeEvent = (
+  e: React.ChangeEvent<HTMLInputElement>,
```
MkMan, line 7: Can you please:
1. define the return type of this function
2. if possible rename `e`. I am generally not a fan of abbreviations. `e` is not so bad when used in the jsx but when used outside it it doesn't have the context and so it  can be confusing.

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+  value: string,
+) => {
+  const modifiedEvent = cloneDeep(e);
```
MkMan, line 10: Would the code below achieve the same thing?

```suggestion
  const modifiedEvent = { ...e };
```

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+  modifiedEvent.target.value = value;
+  return modifiedEvent;
+};
+
+export const pointGen = (pattern: string, num: number) =>
```
MkMan, line 15: Can you please define the return type of this function? 
I am a little bit unclear on the name and what the function does, can you please elaborate?

src/components/PasswordInput/PasswordInput.module.css
```diff
@@ -0,0 +1,30 @@
+@import 'styles/index';
+
+@define-mixin inputFontSize {
```
MkMan, line 3: I was curious as to why you created this mixin 🤔

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+    tempUnmaskedPassword = inputText;
+    tempValue = inputText;
+  } else if (isRemovingChar) {
+    // handle removing chars
+    const diff = inputText.length - value.length;
+    const l = unmaskedPassword.substring(0, selectionEnd);
+    const r = unmaskedPassword.substring(selectionEnd - diff);
+    tempUnmaskedPassword = l + r;
+    tempValue = pointGen(pattern, inputText.length);
+  } else if (propagatedFromPaste !== null) {
```
MkMan, line 101: Is this the same ?
```suggestion
  } else if (propagatedFromPaste) {
```

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+    const { selectionStart, selectionEnd } = cursorRef;
+
+    const l = unmaskedPassword.substring(0, selectionStart);
+    const c = inputText.substring(selectionStart, selectionStart + 1);
+    const r = unmaskedPassword.substring(selectionEnd);
+    tempUnmaskedPassword = l + c + r;
+
+    const pwL = pointGen(pattern, l.length);
+    const pwC = pointGen(pattern, c.length - 1) + c[c.length - 1];
+    const pwR = pointGen(pattern, r.length);
```
MkMan, line 116: I am confused as to what these variables hold, can you please explain them?

src/components/PasswordInput/PasswordInput.types.ts
```diff
@@ -0,0 +1,18 @@
...
+  className?: string;
+  theme?: Themes;
+  onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  label?: string;
+  errorMessage?: string;
+  isEnabled?: boolean;
+  buttonLabels?: {
+    showLabel?: string;
+    hideLabel?: string;
+  };
```
MkMan, line 13: Can you please sort these alphabetically here and in the component destructure?

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+        classNames={textInputClassNames}
+        theme={currentTheme}
+        {...passwordInputProps}
+        errorMessage={errorMessage}
+        isEnabled={isEnabled}
+        label={label}
+        valueControlled={passwordInputProps.value}
+      />
+      <Button
+        aria-label="show/hide password"
```
MkMan, line 48: I think a dynamic aria label would be better here so users know what triggering the button would do. We may need to add a props for the aria label for the show and hide states and fallback on English values if they're missing.

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+    togglePasswordMasking,
+    passwordInputProps,
+  } = useDelayedPasswordMasking({ onChange });
+
+  return (
+    <div className={styles.wrapper}>
+      <TextInputV2
+        classNames={textInputClassNames}
+        theme={currentTheme}
+        {...passwordInputProps}
```
MkMan, line 41: Is there a reason this is in the middle instead of at the end?

src/customHooks/useDelayedPasswordMasking/index.ts
```diff
@@ -0,0 +1 @@
+export { useDelayedPasswordMasking } from './useDelayedPasswordMasking';
```
MkMan, line 1: I was wondering if this hook was likely to be used by anything other than the new component?I am conflicted, should we export it from GEL or just keep with the component where it's used.

Keen to hear everyone's thoughts 😃

MkMan (commented):
Nice work 👏 Good use of TS, and thank you for adding tests where possible 🙏

src/customHooks/useDelayedPasswordMasking/index.ts
cyrusxzw: It looks like only used for `passwordInput`, also hook name is point to `Password`. I would put it in the PasswordInput component.

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+  };
+};
+
+export const onPasteCaptureHandler = (
+  e: React.ClipboardEvent<HTMLInputElement>,
+  unmaskedPassword: string,
+): string => {
+  const { selectionStart, selectionEnd } = e.currentTarget;
+
+  const l = unmaskedPassword.substring(0, selectionStart);
```
cyrusxzw, line 35: What are l,c,r?

src/components/PasswordInput/PasswordInput.module.css
```diff
@@ -0,0 +1,30 @@
...
+  font-size: rem(16px);
+
+  @media (--viewport-md) {
+    font-size: rem(19.2px);
+  }
```
kevin-ho87, line 8: How come there are 2 different font sizes for a button?
`19.2` seems very specific as well.

src/components/PasswordInput/PasswordInput.module.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.button {
+  @mixin inputFontSize;
+
+  position: absolute;
```
kevin-ho87, line 14: Is this wrapped in a `pos relative` container?

src/components/PasswordInput/PasswordInput.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+  // TODO: testing-library/user-event v14 required for onBeforeInput to work
+  it.skip('should mask password by default', () => {
+    render(<PasswordInput label="Password" />);
+    const inputElement = screen.getByRole('textbox', { name: 'Password' });
+
+    userEvent.type(inputElement, 'Aesop123');
+
+    expect(inputElement).toHaveValue('*******3');
+
+    jest.runAllTimers();
```
kevin-ho87, line 38: what does this do? step into the next render sequence?

src/components/PasswordInput/PasswordInput.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+    render(<PasswordInput label="Password" />);
+    const inputElement = screen.getByRole('textbox', { name: 'Password' });
+
+    userEvent.type(inputElement, 'a');
+
+    expect(inputElement).toHaveValue('a');
+
+    act(() => {
+      jest.runAllTimers();
+    });
```
kevin-ho87, line 76: what is the difference of being wrapped in `act()` and not being wrapped like the line above at line 38

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+        errorMessage={errorMessage}
+        isEnabled={isEnabled}
+        label={label}
+        valueControlled={passwordInputProps.value}
+      />
+      <Button
```
kevin-ho87, line 47: does the `button` component come with `type=button` by default?

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
kevin-ho87: I suppose it depends on how nested the object is

src/components/TextInputV2/TextInputV2.tsx
```diff
@@ -60,11 +65,15 @@ const TextInputV2 = forwardRef<HTMLInputElement, TextInputV2Props>(
...
+            onBeforeInput={onBeforeInput}
             onChange={handleOnChange}
+            onCopyCapture={onCopyCapture}
+            onKeyDown={onKeyDown}
+            onPasteCapture={onPasteCapture}
             pattern={pattern}
             ref={ref}
             type={type}
-            value={value}
+            value={valueControlled !== undefined ? valueControlled : value}
```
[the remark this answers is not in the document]
ivandaho: True, forgot about that. And it will allow empty strings in the case of `valueControlled = ''` , so that works well

src/components/TextInputV2/TextInputV2.spec.tsx
```diff
@@ -74,4 +74,44 @@ describe('<TextInputV2 />', () => {
...
   });
+
+  it('should render valueControlled if it is provided', () => {
+    const valueControlled = 'foo';
+    const onKeyDown = jest.fn();
+
+    render(
+      <TextInputV2
+        label="myinput"
+        onKeyDown={onKeyDown}
```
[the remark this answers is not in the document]
ivandaho: Left it in when copy pasting some code, will remove

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMasking.types.ts
```diff
@@ -0,0 +1,28 @@
+export type CursorRef = {
+  selectionStart: number;
+  selectionEnd: number;
+  selectionDirection: 'forward' | 'backward' | 'none';
+};
+
+export type PasswordInputProps = {
```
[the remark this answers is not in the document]
ivandaho: Actually no, I will remove the export statement. Also I realised `PasswordInputProps` is also the name of the type for the actual PasswordInput. Would it be a better idea to rename the one for the hook?

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
ivandaho: Left, Center and Right portions of the new masked/unmasked string. Thoughts on whether these variables should be renamed? Actually I think they look neater when they have the same length. Will a comment or two suffice instead?

```javascript
    // left, right, center of unmasked password
    const l = unmaskedPassword.substring(0, selectionStart);
    const c = inputText.substring(selectionStart, selectionStart + 1);
    const r = unmaskedPassword.substring(selectionEnd);
    tempUnmaskedPassword = l + c + r;

    // left, right, center of masked password
    const pointGenL = pointGen(pattern, l.length);
    const pointGenC = pointGen(pattern, c.length - 1) + c[c.length - 1];
    const pointGenR = pointGen(pattern, r.length);
    tempValue = pointGenL + pointGenC + pointGenR;

```

cyrusxzw: I think just use `const leftPortion or left` rather than abbreviations.

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMasking.ts
```diff
@@ -0,0 +1,142 @@
...
+    setIsMaskingEnabled(!isMaskingEnabled);
+    if (isMaskingEnabled) {
+      setValue(unmaskedPassword);
+    } else {
+      setValue(pointGen(pattern, unmaskedPassword.length));
+    }
+  };
+
+  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const res = onChangeHandler({
```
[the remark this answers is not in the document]
ivandaho: Usually I like to shorten the returned values to `res(ult)`..I think in this case it is not needed, the return values can be initialised in one go

```javascript
    const {
      newValue,
      shouldResetPropagatedFromPaste,
      newUnmaskedPassword,
      cursorRef: newCursorRef,
    } = onChangeHandler({
      ...
    });
```

src/components/PasswordInput/PasswordInput.tsx
ivandaho: Yes, `Button` in `aesop-gel` does

src/components/PasswordInput/PasswordInput.spec.tsx
ivandaho: They run fine without it, I will remove it for now. I think I left it in while trying to see if they were any workarounds for the missing `onBeforeInput` handler in `user-event`

src/components/TextInputV2/TextInputV2.spec.tsx
```diff
@@ -74,4 +74,44 @@ describe('<TextInputV2 />', () => {
...
+        valueControlled={valueControlled}
+      />,
```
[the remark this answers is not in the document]
ivandaho: Will add a few more test cases for `valueControlled`

- should allow empty string as valueControlled

- should prioritize valueControlled over value

src/components/PasswordInput/PasswordInput.spec.tsx
ivandaho: There is a timer to mask the latest inputted char after some delay. For example if the user entered `foo`, without it the string would still be `**o` instead of `foo`

src/components/PasswordInput/PasswordInput.module.css
ivandaho: Thanks, just realised the `wrapper` needs to be have `position: relative` or it will break

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
ivandaho: Agree, to be clearer, I will rename these

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMasking.types.ts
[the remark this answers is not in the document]
MkMan: > Would it be a better idea to rename the one for the hook?

Yeah that sounds good 👍

src/components/PasswordInput/PasswordInput.tsx
ivandaho: only `valueControlled` needs to be at the end. Actually it will pass `value` to `TextInputV2` as well. Would it be clearer if I use a `{...rest}` pattern instead?

```javascript
  const {
    isMaskingEnabled,
    togglePasswordMasking,
    passwordInputProps: { value, ...passwordInputProps },
  } = useDelayedPasswordMasking({ onChange });

  return (
    <div className={styles.wrapper}>
      <TextInputV2
        classNames={textInputClassNames}
        errorMessage={errorMessage}
        isEnabled={isEnabled}
        label={label}
        theme={currentTheme}
        valueControlled={value}
        {...passwordInputProps}
      />

Comment (MkMan) [reply src/components/PasswordInput/PasswordInput.tsx:45]: The snippet you shared above looks good 👌

File: src/components/PasswordInput/PasswordInput.module.css (hunk 5 above)
Comment (andrwchang) [reply src/components/PasswordInput/PasswordInput.module.css:3]: also if we want to create the mixin, I think it is best to make the name more specific - input font could mean anything.

File: src/components/PasswordInput/PasswordInput.module.css (hunk 13 above)
Comment (andrwchang) [reply src/components/PasswordInput/PasswordInput.module.css:8]: possibly to follow the design?

**Review by andrwchang — COMMENTED** [review]
amazing work overall. welldone @ivandaho. also great to see some amazing code review work here

File: src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts (hunk 4 above)
Comment (ivandaho) [reply src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts]: It will return a string consisting of a repeated `pattern` based on the given length. Will refactor to this:

```javascript
/** returns a string consisting of a repeated `pattern` based on a given `length` */
export const pointGen = (pattern: string, length: number): string =>
  length > 0 ? Array(length).fill(pattern).join('') : '';
```

src/components/PasswordInput/PasswordInput.module.css
ivandaho: I was referring to code in `TextInputV2`. However I think this won't be needed here as it is only used in one place (the button)

src/components/PasswordInput/PasswordInput.module.css
ivandaho: Will change this to match design specs:

Mobile (xs to sm): 14px

Anything larger: 16px

src/components/PasswordInput/PasswordInput.spec.tsx
ivandaho: Actually, they still need to be wrapped, or there will be errors in the console, even if the tests are passing. Will add it back.

src/customHooks/useDelayedPasswordMasking/useDelayedPasswordMaskingUtils.ts
MkMan: Always nice to see less lodash 😌

MkMan (approved):
Awesome work 👏 

This PR is probably best closed with a squash merge to avoid the multiple feat commits going into the CHANGELOG

kevin-ho87 (approved):

PR #702, opened 2022-03-28
feat(PrimaryMenu): change primary menu list items from button to hyperlink

masonwongcs:
### Ticket
https://aesoponline.atlassian.net/browse/CON-1186

### Description
To update the Navigation's list item from Button to Hyperlink to improve SEO.

### Screen Grabs
image


src/components/GlobalNavigation/components/DesktopView/components/PrimaryMenu/PrimaryMenu.tsx
```diff
@@ -124,15 +126,17 @@ const PrimaryMenu: PrimaryMenuType = ({ onClose }) => {
             title,
             ...rest
           }) => {
+            const categoryUrl = items[0]?.url;
```
masonwongcs, line 129: Based on the data model in contentful, the first item return from the `graphql` [query](https://github.com/aesop/aesop-graphql-service/blob/develop/src/graphql/resolvers/Query/globalNavigation.js) will be the category. 
```javascript
const collections = contentful.getEntries({
    content_type: 'sectionSiteNavigation',
    include: 3, // Set specifically to avoid extra fetching
  })
    .then((res) => contentful.parseObject(res.items[0]))
    .then((res) => contentful.parseObject(res.categories)
      .filter((cat) => cat.showInMenu || cat.id === 'gift-guide'));
```
image

andrwchang:
dumb question - usually the a link target has default value of `_self`. so the question is why are we adding that attribute? 😃

src/components/GlobalNavigation/components/DesktopView/components/PrimaryMenu/PrimaryMenu.tsx
```diff
@@ -1,5 +1,5 @@
 import cx from 'classnames';
-import React from 'react';
+import React, { SyntheticEvent } from 'react';
```
andrwchang, line 2: 👍 ❤️ 🦸

masonwongcs:
> dumb question - usually the a link target has default value of `_self`. so the question is why are we adding that attribute? 😃

Guess because of this [function](https://github.com/aesop/aesop-gel/blob/dec15d6534ffb6f6b0ee0b52f231937733611d2f/src/components/Hyperlink/Hyperlink.utils.ts#L58)? 
```javascript
const getTargetType = (openInANewWindow: boolean): string =>
  openInANewWindow ? '_blank' : '_self';
```

MkMan (commented):
Is this change ok accessibility wise?
A link is expected to navigate the users and a button is for page interactivity. A button seems like the correct semantic choice given the navigation use case.

cyrusxzw:
Overall, it's great work. Just some minor things:

1. `data-test-ref` if need to change to `NAV_PRIMARY_LINK`
2. Button we have some `aria` to make it more a11y. How about link?

Happy to have a further discussion or demo during our Con Devs Unite.

src/components/GlobalNavigation/components/DesktopView/components/PrimaryMenu/PrimaryMenu.tsx
```diff
@@ -124,15 +126,17 @@ const PrimaryMenu: PrimaryMenuType = ({ onClose }) => {
...
             return (
               <MenuItem
                 dataTestId={`NAV_${id.toUpperCase()}`}
                 dataTestRef="NAV_PRIMARY_BUTTON"
                 id={id}
                 key={id}
                 label={label}
-                onClick={() =>
-                  handleOnOtherCollectionClick(id, {
+                onClick={(event: SyntheticEvent) =>
```
cyrusxzw, line 138: I guess may not need `SyntheticEvent` here? It's already in menuItem.types.ts

cyrusxzw, line 138: I guess we may not need `SyntheticEvent` here, already have it in line 74.

src/components/GlobalNavigation/components/DesktopView/components/PrimaryMenu/PrimaryMenu.tsx
```diff
@@ -150,7 +154,8 @@ const PrimaryMenu: PrimaryMenuType = ({ onClose }) => {
                   })
                 }
                 title={title}
-                type="trigger"
+                type={categoryUrl ? 'link' : 'trigger'}
+                url={categoryUrl}
```
cyrusxzw, line 158: Question: If type is `trigger`, still need `url` attribute?

src/components/GlobalNavigation/components/DesktopView/components/PrimaryMenu/PrimaryMenu.tsx
masonwongcs: I'll update `data-test-ref` to `NAV_PRIMARY_LINK` 👍 
I guess still need to parse in the `SyntheticEvent` here because I do need to prevent the link to be open by default.

src/components/GlobalNavigation/components/DesktopView/components/PrimaryMenu/PrimaryMenu.tsx
masonwongcs: No need I guess, but `url` props are optional here, so it will be fine.

src/components/Hyperlink/Hyperlink.tsx
```diff
@@ -60,7 +61,12 @@ const Hyperlink = forwardRef<HTMLAnchorElement, HyperlinkProps>(
     }
 
     return (
+      // eslint-disable-next-line jsx-a11y/role-supports-aria-props
```
MkMan, line 64: This is a bit of a red flag ⚠️ 😕

Can we please reconsider the approach in this PR? This rule is there for a reason and it looks like we're going to violate accessibility guidelines.

maxlew: Yeah this one is a bit odd, we're trying to move from buttons to hyperlinks for SEO purposes - currently if you view the site with `googlebot` as the UserAgent the navigation is not functional. 

Moving them to hyperlinks solves the SEO problem, but removes the nicety of having these attributes on the Button which are really beneficial to screen readers. From testing voice over on mac, the screen reader correctly lists that it's a collapsible section as a link with these props, obviously it's invalid HTML though so not sure how much it'll work for each screen reader software. 

In short, we're stuck between accessibility and SEO, if you have any thoughts of how we can solve for both we are open to ideas!

MkMan: This is all new to me so I please bear with me. 

I wanted to verify that is the only solution available. I tried accessing YouTube using Safari with the user-agent set to `googlebot` and their menu still worked despite it being made of `<button />`s

https://user-images.githubusercontent.com/33766083/161151105-5f532987-caf3-4e16-b85f-bef4b8264505.mov

When I try to access our website using the same settings the menu doesn't work. It may be that using an `<a />` tag fixes this but it might also be that buttons aren't the root cause of the problem and something else is.

We're trying to get better SEO results (and presumably more money) but we could just as well get sued for having an inaccessible website. I am also concerned that this will start a pattern of breaking best practices instead of resisting and convincing Product Owners that what's being asked is not a good solution. 

In the end this isn't my call to make but if this is the only solution we can find then please don't alter the existing HyperLink component. Instead, please create a copy in the navigation folder to be used for the navigation. That way GEL isn't exporting components that break accessibility guidelines because future users may unknowingly use these new props.

masonwongcs: That sounds like a good idea, I'll create a new component for this.

cyrusxzw (approved):

src/components/Hyperlink/Hyperlink.tsx
masonwongcs: @MkMan @maxlew I've used props spreading in this commit https://github.com/aesop/aesop-gel/pull/702/commits/23729138a0fb92fb825166459c6de0e583d7e54c instead of creating a new component for that, the reason being is that we will have two fairly similar components just with different props. If down the road we have more changes made to the `Hyperlink` component we can still make the changes. 😄

PR #709, opened 2022-04-04
build: run npm ci with --ignore-scripts

MkMan:
From, this [page](https://docs.npmjs.com/cli/v6/commands/npm-install)

> The --ignore-scripts argument will cause npm to not execute any scripts defined in the package.json. 

The aim is to stop running the build when doing an npm install, to speed up the build time.


kevin-ho87 (approved):
Where is the `build` ran normally or otherwise?

MkMan:
> Where is the `build` ran normally or otherwise?

Currently in three places:
1. as part of any npm install or npm ci (because of [this](https://github.com/aesop/aesop-gel/blob/main/package.json#L21) line)
2. in this [workflow step](https://github.com/aesop/aesop-gel/blob/main/.github/workflows/push.yml#L60) to ensure the build has no issues
3. in this [workflow step](https://github.com/aesop/aesop-gel/blob/main/.github/workflows/push.yml#L63) when releasing a new version (I am hoping I can remove this later to reduce the build time further)

PR #715, opened 2022-04-10
build: remove rollup-plugin-dts

MkMan:
The main motive for this change is to reduce the number of build steps to reduce the build time. 

## Current set up
- generate type definitions for the source files and place them in `/dts` 
- use [rollup-plugin-dts](https://www.npmjs.com/package/rollup-plugin-dts) to combine them into one `.d.ts` file (`index.d.ts`)

The outcome looks something like
```ts
// dist/index.d.ts

type ButtonProps = { /* ... */ };
export type Button = VFC<ButtonProps>;

type HyperLinkProps = { /* ... */ };
export type HyperLink = VFC<HyperLinkProps>;
```

## Proposed new set up
- generate the type definitions and place them in `/dist` where the compiled files will go
- do nothing else because the generate index.d.ts file already exports the individual types of files in `/src`

The outcome will look something like
```ts
// dist/index.d.ts

export { Button } from './component/Button/index.d.ts';
export { HyperLink } from './component/HyperLink/index.d.ts';
```

```ts
// dist/components/Button/index.d.ts

type ButtonProps = { /* ... */ };
export type Button = VFC<ButtonProps>;
```


cyrusxzw:
Out of curious, does it reduce the build time in fact?

cyrusxzw (approved):

MkMan:
> Out of curious, does it reduce the build time in fact?

I ran 3 builds to compare and on average it saves at least 10 seconds
|             | **Build 1** | **Build 2** | **Build 3** | Average |
|-------------|-------------|-------------|-------------|---------|
| **Current** | 65.11       | 62.96       | 56.03       | 61.4    |
| **New**     | 49.8        | 50.21       | 51.39       | 50      |

masonwongcs:
Just a random thought, will it reduce or increase bundle sizing in the end?

MkMan:
> Just a random thought, will it reduce or increase bundle sizing in the end?

Good question 🤔 

In term of size shipped to the browser it will be the same because the changes are only to the type definitions.

As for data published to npm below is a comparison of the dist folder right after running build:
- before: 5.8 Mb 
- after: 7.9 Mb

MkMan:
Found issues on Web UI when attempted using this build. Switching to Draft until it can be solved.

PR #716, opened 2022-04-11
Enhance checkbox

MkMan:
Changes
- add a prop to allow styling the checkbox wrapper div
- relax the label prop type to allow it to be jsx. This is for a use case where the label is Contentful Rich text


src/components/Checkbox/Checkbox.types.ts
```diff
@@ -11,11 +11,12 @@ type CheckboxProps = {
   id?: string;
   isChecked?: boolean;
   isEnabled?: boolean;
-  label: string;
+  label: ReactChild;
```
cyrusxzw, line 14: If need string as an option?  ` label: ReactChild | string`;

MkMan: `ReactChild` already includes `string` 😃 

image

cyrusxzw: wow, nice

cyrusxzw (approved):

ivandaho (approved):

masonwongcs (approved):

PR #721, opened 2022-04-20
fix(con-1286): fix carousel arrow not centered

cyrusxzw:
Fix carousel arrow issue and update snapshots. You may ignore snapshots changes, just focus on `NextButton.module.css` and `PreviousButton.module.css`.

Before:
Screen Shot 2022-04-20 at 12 15 34 pm

After:
Screen Shot 2022-04-20 at 12 15 55 pm


src/components/Button/Button.tsx
```diff
@@ -68,11 +68,13 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(function ButtonRef(
       type={type}
       {...otherProps}
     >
-      <Loading
-        isLoading={isLoading}
-        screenReaderText={loadingScreenReaderText}
-        theme={isAlternate ? getOppositeTheme(theme) : theme}
-      />
+      {isLoading && (
```
MkMan, line 71: I think this breaks accessibility. From memory the <Loading> component should always be there and it will update a div with role alert if the isLoading prop is true.

cyrusxzw: Yea, if consider about accessibility, I think we can add another prop `hasLoadingPlaceholder`, cause loading placeholder should not use for every button, right?

MkMan: How will that prop be different to the `isLoading` prop?

cyrusxzw: Update.

MkMan (commented):
I hate snapshot testing but was deleting the files intentional 🤔 😄 ?

MkMan (approved):

src/components/Carousel/components/NextButton/NextButton.module.css
```diff
@@ -20,6 +20,10 @@
     transform: translate3d(80px, 0, 0);
   }
 
+  span[class*='Loading'] {
+    width: auto;
+  }
```
kevin-ho87, line 25: what element is this?

cyrusxzw: Screen Shot 2022-04-20 at 1 39 40 pm

kevin-ho87: this width auto update makes the icon centered?

cyrusxzw: Yes, previous width is 100%

kevin-ho87 (approved):

PR #724, opened 2022-04-22
feat(Hyperlink): set url in Hyperlink to optional

ivandaho:
`url` prop in Hyperlink is only used to provide `href` to `a`
`a.href` is optional, so this should be consistent


src/components/Hyperlink/Hyperlink.types.ts
```diff
@@ -33,7 +33,7 @@ type HyperlinkProps = {
   textAlign?: TextAlign;
   theme?: Themes;
   title?: string;
-  url: string;
+  url?: string;
```
MkMan, line 36: Please don't apply this change, links without a `href` are invalid and inaccessible

MkMan: https://user-images.githubusercontent.com/33766083/164605584-d013697a-cc46-4acc-9b1b-92099b5ee5d4.mov

PR #725, opened 2022-04-26
fix(Button): remove loading placeholder from layout when not loading

MkMan:
This change removes the Loading component from layout to fix issues where it was messing up layout for siblings.

## Before
image

## After
image


src/components/Button/Button.module.css
```diff
@@ -121,3 +121,8 @@
     }
   }
 }
+
+.hideLoadingIndicator {
```
cyrusxzw, line 125: How about using CSS Clipping?

 ```
.hideLoadingIndicator {
    border: 0;
    clip: rect(1px 1px 1px 1px); /* IE6, IE7 */
    clip; rect(1px, 1px, 1px, 1px);
    height: 1px;
    margin: -1px;
    overflow: hidden;
    padding: 0;
    position: absolute;
    width: 1px; 
} 
```

cyrusxzw:
Good solution by adding a hidden className. Could you help me to check if it works well in carousel component without my fix?

src/components/Button/Button.module.css
MkMan: I haven't tried it but I am keen to understand how it's better than `left: -999px`

cyrusxzw: I just feel like `left: -999px` is a bit of a hack. 😄

MkMan: Fair enough 👍  I'll see if this approach has an impact on a11y and give it a shot

cyrusxzw: My refs: https://www.nomensa.com/blog/how-improve-web-accessibility-hiding-elements

MkMan: I used the css-clip example shown in this [page](https://webaim.org/techniques/css/invisiblecontent/) which is referenced on the page you linked 😃 
Thanks for the suggestion

MkMan:
> Good solution by adding a hidden className. Could you help me to check if it works well in carousel component without my fix?

Yes it works without the fix 👌 

image

image

Would you like me to revert it?

cyrusxzw:
> > Good solution by adding a hidden className. Could you help me to check if it works well in carousel component without my fix?
> 
> Yes it works without the fix 👌
> 
> image
> 
> image
> 
> Would you like me to revert it?

Yes please, thanks!

cyrusxzw (approved):

PR #726, opened 2022-05-04
fix(Image): refine Image TS `type` property

MkMan:
While using this component I noticed that the `type` prop was a `string`. After digging into the code I found it's only used on those 2 lines and can be made more specific 

https://github.com/aesop/aesop-gel/blob/c8f6109a14bb18d8838612386f9b978cd5b0b248/src/components/Image/Image.tsx#L25-L26


src/components/Image/Image.types.ts
```diff
@@ -24,7 +24,7 @@ type ImageProps = {
     xSmall?: string;
   };
   style?: CSSProperties;
-  type?: string;
+  type?: 'componentFullBleedImage' | 'componentSquareImage';
```
masonwongcs, line 27: Just curious will there be any other image type in the future?

masonwongcs (approved):

cyrusxzw (approved):

src/components/Image/Image.types.ts
MkMan: I am not sure 😅  at the moment these were the only 2 I could find in the code. Did you have something in mind?

masonwongcs: I think we can keep it as these 2 for now 😄

PR #729, opened 2022-05-09
feat(con-1265): add props and update style of carousel 

cyrusxzw:
In order to show more fields, such as price/size and product badge, I added some optional props, then we are able to render different layout based off `carouselType`. 

Changes in this PR include:

- Add more props
- Styles update to match design
- Change slide setting to show half product (see picture below)
- Update snapshots

##Screenshot
Screen Shot 2022-05-11 at 9 11 45 am


src/components/Carousel/Carousel.module.css
```diff
@@ -23,6 +23,18 @@
...
+    .commerce & {
+      height: auto;
+
+      @media (--viewport-md) {
+        height: auto;
+      }
+
+      @media (--viewport-lg) {
+        height: auto;
+      }
```
MkMan, line 35: Are these needed if height is already set to auto on line 27?

src/components/Carousel/Carousel.types.ts
```diff
@@ -6,17 +6,21 @@ import type { ImageProps } from '../Image/Image.types';
...
   isLoading?: boolean;
   image: ImageProps;
+  price?: string;
+  size?: string;
   url?: string;
 };
 
 type CarouselProps = {
   autoplaySpeed?: number;
+  carouselType?: 'content' | 'commerce';
```
MkMan, line 23: I noticed this property type is repeated across 4 files (in one of them it's a string). Is it better to define it somewhere and shared it across the components?

src/components/Carousel/Carousel.types.ts
```diff
@@ -6,17 +6,21 @@ import type { ImageProps } from '../Image/Image.types';
...
 type Slide = {
+  badge?: string;
   caption?: string;
   description?: string;
   heading?: string;
   id?: string;
   isLoading?: boolean;
   image: ImageProps;
+  price?: string;
+  size?: string;
```
MkMan, line 17: Looking at the other properties in this type they seem generic (not specific to a product). They describe what each property does in the UI.

image


I noticed from the screenshot you attached that these properties are place under the `heading`. Perhaps they can be `subheading` instead?

src/components/Carousel/Carousel.module.css
cyrusxzw: right, don't need.

src/components/Carousel/Carousel.types.ts
cyrusxzw: I planed to do that, but I found there already is `CarouselType` defined on line 40, I don't know how to name it.

src/components/Carousel/Carousel.types.ts
cyrusxzw: These props will only used to product and show size and price only, so I think price&size is more accuracy?

src/components/Carousel/Carousel.types.ts
MkMan: Ah I see 🤔 I guess you can either rename one of them. CarouselType can be something like `CarouselComponentType` or rename this prop (perhaps `carouselKind` ?)

cyrusxzw: `CarouselComponentType` sounds good, thanks

src/components/Carousel/Carousel.types.ts
MkMan: I guess that's also related. Does a carousel need to know that it's showing `content` or `commerce `?
A carousel is just a presentational component and doesn't need to know what its content is. Ideally, its props should not be tied to the content it displays. 

Consider in the future a dev tries to use this component and sees that each slide can take `price` and `size`. That doesn't immediately make it clear what those values do or where they will be displayed. They will need to look into the code to figure out how to use them. In contrast, the other properties in the Slide type give a general idea about their role in the user interface.

Keen to hear your thoughts 🙂

cyrusxzw: It make sense, it'll change it to `subheading`.

src/components/Carousel/Carousel.module.css
```diff
@@ -36,13 +40,33 @@
     display: flex;
   }
 
+  :global(.slick-slide) {
+    & > div {
+      height: 100%;
+    }
+  }
+
   :global(.slick-list) {
```
kevin-ho87, line 49: what does `:global` do?

cyrusxzw: It declares a global classname without hash. I didn't write it. 😄

src/components/Carousel/Carousel.utils.tsx
```diff
@@ -80,24 +86,24 @@ const getCarouselSettings = ({
     {
       breakpoint: BREAKPOINTS.get('sm').maxWidth,
       settings: {
-        slidesToShow: 1,
+        slidesToShow: isCommerceCarousel ? 1.5 : 1,
```
kevin-ho87, line 89: `1.5` slides to show?

src/components/Carousel/components/CarouselIntroduction/CarouselIntroduction.module.css
```diff
@@ -17,6 +25,16 @@
   margin-top: 0;
   margin-bottom: 9px;
 
+  .commerce & {
+    margin-bottom: 0;
+
+    > h2 {
```
kevin-ho87, line 31: does this heading have a className?

src/components/Carousel/Carousel.utils.tsx
cyrusxzw: Yes.
Screen Shot 2022-05-11 at 2 14 46 pm

src/components/Carousel/components/CarouselIntroduction/CarouselIntroduction.module.css
cyrusxzw: they are all called `heading`

src/components/Carousel/Carousel.utils.tsx
kevin-ho87: ah i see

src/components/Carousel/components/CarouselIntroduction/CarouselIntroduction.module.css
kevin-ho87: If possible are you able to add className and target instead of "> h2".

cyrusxzw: It used `SectionHeading`, and not able to pass a classname. 😭

MkMan:
I think it would be good to move away from using `carouselType` in a dumb component such as this to avoid adding business logic to it. It would be good to look at the component and see what parts of it we wish to change to facilitate the UX request and then expose props that make sense to this component that allow the change.

Let’s consider a simpler example. We have the GEL Button and let’s say that on a PDP page we want that button to be **red**, on the Account page we want it **blue** and for other places we want it **black**. We can go about it in 2 ways

## 1. Make the Button know what page it’s on and select the colour

``` tsx
// GEL
type Props = {
  label: string;
  // ...
  page: 'pdp' | 'account' | 'default';
};

const Button = ({ label, page }) => {
  const colourMap = {
    pdp: 'red',
    account: 'blue',
    default: 'black'
  };

  return <button style={{ color: colourMap[page] }}>{label}</button>
};

// Web UI
<Button label="Add to cart" page="pdp" />
<Button label="Log out" page="account" /> 
```

This approach is not ideal because a dumb, presentational component like a button now has logic inside it related to factors outside the button (i,e, the page). If in the future we wish to change those colours or add more variations we will need to update: 1. GEL, 2. Web UI

But what we really want to do is have control over the colour of the button. The choice of what the colour should be can be determined by the app using the button. This is why I think the second approach is better.

## 2. Allow control of the UI through props

```tsx
// GEL
type Props = {
  label: string;
  // ...
  color: string;
};

const Button = ({ label, color }) => {
  return <button style={{ color }}>{label}</button>
};

// Web UI
<Button label="Add to cart" color="red" />
<Button label="Log out" page="blue" />
```

This keeps business logic outside of our low-level dumb components and in the middle layers (gql, UI apps). It also makes updating the colour easier as we just do that in 1 place (Web UI) instead of 2 places.

cyrusxzw:
> I think it would be good to move away from using `carouselType` in a dumb component such as this to avoid adding business logic to it. It would be good to look at the component and see what parts of it we wish to change to facilitate the UX request and then expose props that make sense to this component that allow the change.
> 
> Let’s consider a simpler example. We have the GEL Button and let’s say that on a PDP page we want that button to be **red**, on the Account page we want it **blue** and for other places we want it **black**. We can go about it in 2 ways
> ## 1. Make the Button know what page it’s on and select the colour
> 
> ```tsx
> // GEL
> type Props = {
>   label: string;
>   // ...
>   page: 'pdp' | 'account' | 'default';
> };
> 
> const Button = ({ label, page }) => {
>   const colourMap = {
>     pdp: 'red',
>     account: 'blue',
>     default: 'black'
>   };
> 
>   return <button style={{ color: colourMap[page] }}>{label}</button>
> };
> 
> // Web UI
> <Button label="Add to cart" page="pdp" />
> <Button label="Log out" page="account" /> 
> ```
> 
> This approach is not ideal because a dumb, presentational component like a button now has logic inside it related to factors outside the button (i,e, the page). If in the future we wish to change those colours or add more variations we will need to update: 1. GEL, 2. Web UI
> 
> But what we really want to do is have control over the colour of the button. The choice of what the colour should be can be determined by the app using the button. This is why I think the second approach is better.
> ## 2. Allow control of the UI through props
> 
> ```tsx
> // GEL
> type Props = {
>   label: string;
>   // ...
>   color: string;
> };
> 
> const Button = ({ label, color }) => {
>   return <button style={{ color }}>{label}</button>
> };
> 
> // Web UI
> <Button label="Add to cart" color="red" />
> <Button label="Log out" page="blue" />
> ```
> 
> This keeps business logic outside of our low-level dumb components and in the middle layers (gql, UI apps). It also makes updating the colour easier as we just do that in 1 place (Web UI) instead of 2 places.

Thanks for suggestion, but I don't think `Button` is a suitable example. Firstly, button is using by lots of components and pages as a low level (children) UI component, so we don't want to add some props based off page. Second, `Carousel` is still a dumb component, we planned to add a new component called `productCarousel`, but after compared with design, I found there are only minor changes and 90% of codes will not be changed. So we just updated on the existing carousel component. It's mostly like `TextInput` and `TextInputV2`. Instead of having 2 very similar components, we combine them together by passing a prop `carouselType`. That means we have 2 dumb components (common carousel and commence carousel) basically. Last, we also have lots of styles changes based off `carouselType`. It's not only just change color, but font, margin etc. By checking `isCommerceCarousel`, new className will be assigned to the carousel. If we don't have a type to distinguish, how to change the style?

JasonPingAesop:
> > I think it would be good to move away from using `carouselType` in a dumb component such as this to avoid adding business logic to it. It would be good to look at the component and see what parts of it we wish to change to facilitate the UX request and then expose props that make sense to this component that allow the change.
> > Let’s consider a simpler example. We have the GEL Button and let’s say that on a PDP page we want that button to be **red**, on the Account page we want it **blue** and for other places we want it **black**. We can go about it in 2 ways
> > ## 1. Make the Button know what page it’s on and select the colour
> > ```tsx
> > // GEL
> > type Props = {
> >   label: string;
> >   // ...
> >   page: 'pdp' | 'account' | 'default';
> > };
> > 
> > const Button = ({ label, page }) => {
> >   const colourMap = {
> >     pdp: 'red',
> >     account: 'blue',
> >     default: 'black'
> >   };
> > 
> >   return <button style={{ color: colourMap[page] }}>{label}</button>
> > };
> > 
> > // Web UI
> > <Button label="Add to cart" page="pdp" />
> > <Button label="Log out" page="account" /> 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This approach is not ideal because a dumb, presentational component like a button now has logic inside it related to factors outside the button (i,e, the page). If in the future we wish to change those colours or add more variations we will need to update: 1. GEL, 2. Web UI
> > But what we really want to do is have control over the colour of the button. The choice of what the colour should be can be determined by the app using the button. This is why I think the second approach is better.
> > ## 2. Allow control of the UI through props
> > ```tsx
> > // GEL
> > type Props = {
> >   label: string;
> >   // ...
> >   color: string;
> > };
> > 
> > const Button = ({ label, color }) => {
> >   return <button style={{ color }}>{label}</button>
> > };
> > 
> > // Web UI
> > <Button label="Add to cart" color="red" />
> > <Button label="Log out" page="blue" />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This keeps business logic outside of our low-level dumb components and in the middle layers (gql, UI apps). It also makes updating the colour easier as we just do that in 1 place (Web UI) instead of 2 places.
> 
> Thanks for suggestion, but I don't think `Button` is a suitable example. Firstly, button is using by lots of components and pages as a low level (children) UI component, so we don't want to add some props based off page. Second, `Carousel` is still a dumb component, we planned to add a new component called `productCarousel`, but after compared with design, I found there are only minor changes and 90% of codes will not be changed. So we just updated on the existing carousel component. It's mostly like `TextInput` and `TextInputV2`. Instead of having 2 very similar components, we combine them together by passing a prop `carouselType`. That means we have 2 dumb components (common carousel and commence carousel) basically. Last, we also have lots of styles changes based off `carouselType`. It's not only just change color, but font, margin etc. By checking `isCommerceCarousel`, new className will be assigned to the carousel. If we don't have a type to distinguish, how to change the style?

From my understanding, we need check what's actually difference between old and new carousel. If just style difference, even it has color, margin, padding, etc.. change, we should not add style switch logic inside component. Instead, we should allow pass style as prop into the component to change the style. We can build two css files for old and new carousel respectively, and make them exportable, so when we actually want to switch them, we do that in web-ui like below:

```
// web-ui
// for Old carousel
import {oldStyle} from '@aesop-enginneer/aesop-gel/components/Carousel'
 
<Carousel style={oldStyle} />


// for New carousel
import {newStyle} from '@aesop-enginneer/aesop-gel/components/Carousel'

<Carousel style={newStyle} />


// for another style carousel

import {myStyle} from '../../some-places'

<Carousel style={myStyle} />

```
If there is a huge structure difference, not just heading is hidden (can can control this in style too), but more like sections reorder (content at top, heading at bottom), we can create another component as most tag will be different, but let web-ui to determine which to use rather than pass page specify prop to switch them.

cyrusxzw:
> > > I think it would be good to move away from using `carouselType` in a dumb component such as this to avoid adding business logic to it. It would be good to look at the component and see what parts of it we wish to change to facilitate the UX request and then expose props that make sense to this component that allow the change.
> > > Let’s consider a simpler example. We have the GEL Button and let’s say that on a PDP page we want that button to be **red**, on the Account page we want it **blue** and for other places we want it **black**. We can go about it in 2 ways
> > > ## 1. Make the Button know what page it’s on and select the colour
> > > ```tsx
> > > // GEL
> > > type Props = {
> > >   label: string;
> > >   // ...
> > >   page: 'pdp' | 'account' | 'default';
> > > };
> > > 
> > > const Button = ({ label, page }) => {
> > >   const colourMap = {
> > >     pdp: 'red',
> > >     account: 'blue',
> > >     default: 'black'
> > >   };
> > > 
> > >   return <button style={{ color: colourMap[page] }}>{label}</button>
> > > };
> > > 
> > > // Web UI
> > > <Button label="Add to cart" page="pdp" />
> > > <Button label="Log out" page="account" /> 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This approach is not ideal because a dumb, presentational component like a button now has logic inside it related to factors outside the button (i,e, the page). If in the future we wish to change those colours or add more variations we will need to update: 1. GEL, 2. Web UI
> > > But what we really want to do is have control over the colour of the button. The choice of what the colour should be can be determined by the app using the button. This is why I think the second approach is better.
> > > ## 2. Allow control of the UI through props
> > > ```tsx
> > > // GEL
> > > type Props = {
> > >   label: string;
> > >   // ...
> > >   color: string;
> > > };
> > > 
> > > const Button = ({ label, color }) => {
> > >   return <button style={{ color }}>{label}</button>
> > > };
> > > 
> > > // Web UI
> > > <Button label="Add to cart" color="red" />
> > > <Button label="Log out" page="blue" />
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This keeps business logic outside of our low-level dumb components and in the middle layers (gql, UI apps). It also makes updating the colour easier as we just do that in 1 place (Web UI) instead of 2 places.
> > 
> > 
> > Thanks for suggestion, but I don't think `Button` is a suitable example. Firstly, button is using by lots of components and pages as a low level (children) UI component, so we don't want to add some props based off page. Second, `Carousel` is still a dumb component, we planned to add a new component called `productCarousel`, but after compared with design, I found there are only minor changes and 90% of codes will not be changed. So we just updated on the existing carousel component. It's mostly like `TextInput` and `TextInputV2`. Instead of having 2 very similar components, we combine them together by passing a prop `carouselType`. That means we have 2 dumb components (common carousel and commence carousel) basically. Last, we also have lots of styles changes based off `carouselType`. It's not only just change color, but font, margin etc. By checking `isCommerceCarousel`, new className will be assigned to the carousel. If we don't have a type to distinguish, how to change the style?
> 
> From my understanding, we need check what's actually difference between old and new carousel. If just style difference, even it has color, margin, padding, etc.. change, we should not add style switch logic inside component. Instead, we should allow pass style as prop into the component to change the style. We can build two css files for old and new carousel respectively, and make them exportable, so when we actually want to switch them, we do that in web-ui like below:
> 
> ```
> // web-ui
> // for Old carousel
> import {oldStyle} from '@aesop-enginneer/aesop-gel/components/Carousel'
>  
> <Carousel style={oldStyle} />
> 
> 
> // for New carousel
> import {newStyle} from '@aesop-enginneer/aesop-gel/components/Carousel'
> 
> <Carousel style={newStyle} />
> 
> 
> // for another style carousel
> 
> import {myStyle} from '../../some-places'
> 
> <Carousel style={myStyle} />
> ```
> 
> If there is a huge structure difference, not just heading is hidden (can control this in style too), but more like sections reorder (content at top, heading at bottom), we can create another component as most tag will be different, but let web-ui to determine which to use rather than pass page specify prop to switch them.

It has both style and structure changes, but all of them are really tiny. This is why I don't want to create a new component with high similarity.

JasonPingAesop (approved):

kevin-ho87 (approved):

PR #730, opened 2022-05-10
fix(errors): make message colour accessible

MkMan:


MkMan:
image

Colour not accessible on dark background. Waiting for feedback from UX team.

PR #738, opened 2022-05-20
feat(Modal): enhance to take extra children

MkMan:
This allows the Modal to some JSX and place it in the top left corner opposite the close button.

I am reluctant to merge this as I couldn't find a clean way to get the desired layout without changing the current spacings.

## Before
image

image

## After
image

image


src/components/Modal/components/ModalBody/ModalBody.module.css
```diff
@@ -89,3 +101,12 @@
     outline-offset: 2px;
   }
 }
+
+.modalContent {
+  padding: 0 var(--layout-sm-margin) var(--layout-sm-margin);
+
```
ivandaho, line 107: Would it be acceptable not to change the padding of the top row buttons? For example the login modal design specs has the title lower than GEL's modal's default, and we implemented that by [forcing a 40px padding](https://github.com/aesop/aesop-com/blob/main/libs/login-modal/src/styles/index.scss#L6).

We can add a negative `margin-top` here (looks to be about 30px) to default it to the same spacing as before these changes
And if the `topLeftChildren` or the top row in general needs more space, the developer would override this margin as needed according to the design specs where `Modal` is used

PR #739, opened 2022-05-20
feat(Modal): remove modal max height 

MkMan:
These changes are to remove the max-height restriction on the modal (as per the UX team's request).

## Before
before

## After
after


src/components/Modal/Modal.stories.mdx
```diff
@@ -3,7 +3,7 @@ import * as knobs from '@storybook/addon-knobs';
 import { State } from '~.storybook/State';
 import { Button } from '~/components/Button';
 import { Modal } from './Modal';
-import { ModalFixture } from './Modal.fixture';
+import { ModalFixture } from './Modal.fixture.tsx';
```
ivandaho, line 6: Is the .tsx extension needed?

MkMan: I could have swore that Storybook complained when it wasn't there but I just removed it and it works fine. I guess it's not needed, I'll remove it. Thank you 😃

MkMan: Done ✔️

cyrusxzw:
Looks good, but may need to check the component using Modal, make sure they are still working well. 😄

cyrusxzw (approved):

PR #741, opened 2022-05-24
Enhance modal

MkMan:
## Changes
1. Allow component to take a composition in the top left position opposite the Close button
2. Remove the modal max height restriction on Tablet/Desktop views 
3. Refactor to remove React Fragment containing only 1 child

## Before

https://user-images.githubusercontent.com/33766083/170142715-27899c4d-b8e2-4d39-ad45-08cb4552298f.mov

## After

https://user-images.githubusercontent.com/33766083/170142789-c4924ecd-d583-410c-b58d-44cf22e4810b.mov

## With a `topLeftChildren`

image


src/components/Modal/components/ModalBody/ModalBody.tsx
```diff
@@ -12,23 +12,25 @@ const ModalBody: ModalBodyType = ({
...
-    { [styles.isVisible]: isVisible },
     styles[currentTheme],
     className,
   );
 
   return (
     <div className={classSet}>
       {children}
+      {topLeftChildren && (
+        <div className={styles.topLeft}>{topLeftChildren}</div>
```
kevin-ho87, line 32: I am guessing any styling to move this element further would come from the child?

kevin-ho87: I am not sure if the styling of this top left holder should be done here as the element is unknown.

src/components/Modal/Modal.tsx
```diff
@@ -38,48 +39,49 @@ const Modal: ModalType = ({
 
   const isMobile = isViewport('xs to sm only');
 
-  return (
+  return createPortal(
     <>
```
kevin-ho87, line 43: Is this update to remove the previous standalone `<>` ?

MkMan: Yeah, previously it was 

```tsx
return (<>
  createPortatl(...)
</>)
```

so I just made it
```tsx
return createPortatl(...)
```

src/components/Modal/components/ModalBody/ModalBody.tsx
MkMan: > I am guessing any styling to move this element further would come from the child?

Yeah

> I am not sure if the styling of this top left holder should be done here as the element is unknown.

I get what you mean but literally [all the styling](I am not sure if the styling of this top left holder should be done here as the element is unknown.) in there is to position the element in the top left in way similar to the close button. The reason I favoured putting that here is because dimensions used to position the close button are here so I figured this is a good place to make sure they are in sync.

Do you see that being an issue?

kevin-ho87: no, it should be ok

kevin-ho87 (approved):

cyrusxzw:
I saw the heading is slightly covered by close icon on mobile view. Does it need to be considered?

ivandaho (approved):

MkMan:
> I saw the heading is slightly covered by close icon on mobile view. Does it need to be considered?

I noticed it too but it's existing behaviour. I think it's up to the user of the Modal to assign spacing to the top of their content so it doesn't clash with the close icon.

cyrusxzw (approved):

PR #746, opened 2022-06-07
feat(con-1391): build table

cyrusxzw:
This table supports both normal table and contentful table usage.

1. Normal table:
Passing Columns and Rows as props in following structure:

```
Columns = [
  { key: 'id', title: 'ID' },
  { key: 'name', title: 'Name' },
  { key: 'age', title: 'Age' },
];

```
```
Rows = [
  { id: 1, name: 'John', age: 20 },
  { id: 2, name: 'Mike', age: 30 },
  { id: 3, name: 'Steven', age: 25 },
  { id: 4, name: 'Brian', age: 40 },
];
```

2. Contentful table:
Cause we use `rich-text-html-renderer ` for contentful table, it comes with html structure. For example:
```

<tr>
  <td>Item 1</td>
</tr>
<tr>
  <td>Item 2</td>
</tr>
<tr>
  <td>Item 3</td>
</tr>
```
Contentful table as `children` passed into Table component.


src/components/Table/Table.types.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+type ColumnsProps = {
+  columns?: Column[];
+};
+
+type RowsProps = {
+  columns?: Column[];
+  rows?: any[];
+};
```
MkMan, line 31: Is it possible to define the types currently set as `any`?

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,23 @@
...
+const Table = forwardRef<HTMLTableElement, TableProps>(
+  ({ caption, children, className, columns, dataTestRef, rows }, ref) => {
+    const baseClassSet = cx(styles.base, className);
+
+    return (
+      <table className={baseClassSet} data-test-ref={dataTestRef} ref={ref}>
+        {caption && <caption>{caption}</caption>}
+        <Columns columns={columns} />
+        <Rows columns={columns} rows={rows} />
+        <tbody>{children}</tbody>
```
MkMan, line 17: Is it possible to have both `rows` and `children`? That would cause there to be two `tbody` right?

src/components/Table/Table.stories.mdx
```diff
@@ -0,0 +1,18 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs';
+import * as knobs from '@storybook/addon-knobs';
```
MkMan, line 2: Can you please remove this import? The knobs addon is deprecated

src/components/Table/Table.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+  it('should render base component correctly', () => {
+    const { container } = render(
+      <Table dataTestRef="test-data-ref">
+        <tr>
+          <td>test</td>
+        </tr>
+      </Table>,
+    );
+
+    expect(container).toMatchSnapshot();
```
MkMan, line 16: Can you please replace this test with actual tests of the component functionality?

src/components/Table/Table.tsx
kevin-ho87: what is the intended markup for this table component? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table

src/components/Table/Table.types.ts
cyrusxzw: Nope, rows type based off headings. It should be an array, but uncertain object (both key and type).

src/components/Table/Table.tsx
cyrusxzw: Cause it need to support both normal table and contentful table. For contentful table usage, it will not have 2 tbody. But for normal table, if someone passing `rows` and `children`, it will have 2 tbody. So have to avoid passing both.

src/components/Table/Table.stories.mdx
cyrusxzw: Oh yeah, I'll remove.

src/components/Table/Table.spec.tsx
cyrusxzw: Sure, I'll add some tests.

src/components/Table/Table.tsx
cyrusxzw: Screen Shot 2022-06-08 at 9 57 45 am

MkMan: Try this type, hopefully it should prevent the component from being used with both

```ts
type CommonTableProps = {
  caption?: string;
  className?: string;
  columns?: Column[];
  dataTestRef?: string;
};

type TableProps = CommonTableProps &
  (
    | {
        children: ReactNode;
        rows?: never;
      }
    | {
        children?: never;
        rows?: any[];
      }
  );
```

src/components/Table/Table.types.ts
MkMan: I didn't get what you meant sorry, can you please elaborate?

src/components/Table/Table.tsx
cyrusxzw: I'll remove tbody of children, same reason [here](https://github.com/contentful/rich-text/pull/269).

src/components/Table/Table.types.ts
cyrusxzw: For example: 
```
Columns = [
    {
        key: 'name',
        title: 'name'
    },
    {
         key: 'age',
         title: 'age'
     }
]
```
then Row type is ```
{
    name: string,
    age: number
}
```
   but if ```
Columns = [
    {
        key: 'id',
        title: 'id'
    },
    {
         key: 'name',
         title: 'name'
     }
]
```
Row type will be ```
{
    id: string,
    name: string
}

src/components/Table/Table.tsx
MkMan: Can the component be used with both `children` and `rows`?

cyrusxzw: Better not, I just updated.

cyrusxzw: Done

masonwongcs (approved):

PR #749, opened 2022-06-08
feat: allow child radio elem to be exported

kevin-ho87:
Allow Radio element to be exported via `<RadioGroup.Radio {...props} />`

image


ivandaho (approved):

src/components/RadioGroup/RadioGroup.stories.mdx
```diff
@@ -39,9 +41,23 @@ export const Template = (props) => (
       name: 'radioGroup',
       options: RadioGroupFixture.options,
     }}
+    argTypes={{ theme: { control: 'radio', options: ['dark', 'light'] } }}
```
MkMan, line 44: Do you need to manually specify this? It worked out of the box for the Button component 

image

src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -16,6 +29,9 @@ type RadioGroupProps = {
   value?: string;
 };
 
-type RadioGroupType = ComponentWithoutChildren<RadioGroupProps>;
+type RadioType = React.VFC<RadioProps>;
```
MkMan, line 32: A while back we made a decision not to use `FC` or `VFC` in GEL. Would `ComponentWithoutChildren` work in place of `VFC` here?

MkMan (approved):
Nice work 👌 Is this change prep work for something upcoming?

kevin-ho87:
> Nice work 👌 Is this change prep work for something upcoming?

Yeah, there is this vertical type modal radio that needs to have an element in between
image

src/components/RadioGroup/RadioGroup.stories.mdx
kevin-ho87: It is a weird one. It originally did automatically generate all the props until I extracted the Radio component. It could be something due to that.

MkMan: Ah fair enough 👍

src/components/RadioGroup/RadioGroup.types.ts
kevin-ho87: updated.

PR #758, opened 2022-07-26
feat(con-1442): add carousel in product card

cyrusxzw:
Add a carousel in the existing product grid item. By checking  `hasCarousel` prop, we are able to control show an image or carousel in contentful.

https://user-images.githubusercontent.com/29698458/181652203-7aa15693-6ebf-4a95-8852-314575b7c284.mov


src/components/ProductGridItem/Carousel/NextButton/__snapshots__/NextButton.spec.tsx.snap
```diff
@@ -0,0 +1,41 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Carousel.NextButton /> should render base component correctly 1`] = `
```
JasonPingAesop, line 3: shall we not use snapshot as testing?

src/components/ProductGridItem/Carousel/PreviousButton/PreviousButton.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+  it('should render base component correctly', () => {
+    const { container } = render(
+      <PreviousButton
+        onClick={() => {
+          return;
+        }}
+      />,
+    );
+
+    expect(container).toMatchSnapshot();
```
JasonPingAesop, line 16: same, remove snapshot?

src/components/ProductGridItem/Carousel/PreviousButton/PreviousButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+import styles from './PreviousButton.module.css';
+import type { PreviousButtonType } from './PreviousButton.types';
+
+const PreviousButton: PreviousButtonType = ({ onClick, theme }) => (
+  <Button
+    className={styles.base}
+    isInline={true}
+    onClick={onClick}
+    theme={theme}
+    title="previous slide"
```
JasonPingAesop, line 15: Use props to set this title rather than hardcode it? Just think of locale

cyrusxzw: We are using the same way on the existing carousel, not sure if we have the copy in CF.

masonwongcs (approved):

PR #759, opened 2022-07-27
feat(CON-1461): fix line height alignment issue of desktop navigation

masonwongcs:
## Ticket
https://aesoponline.atlassian.net/browse/CON-1461

## Description
- To align the Stores button with the rest of the navigation.

## Task Summary
- It is because `a` is an inline element and `button` was a block element in the header in this scenario, the default `line-height` value will affect the alignment of the nav
- Reset the `line-height` to the initial value

## Screen Grabs
Before and After
Untitled


src/components/GlobalNavigation/components/DesktopView/components/MenuItem/MenuItem.module.css
```diff
@@ -20,6 +20,10 @@
   &:last-child {
     margin-right: 0;
   }
+
+  > button {
```
cyrusxzw, line 24: Is there a classname?

masonwongcs: Because the style was inherited from parent element, I have to use a children selector.

kevin-ho87:
I cant see the difference between before and after. Maybe i am just getting old 😅

masonwongcs:
> I cant see the difference between before and after. Maybe i am just getting old 😅

@kevin-ho87 there's a ~1px difference though 😄

cyrusxzw (approved):

kevin-ho87 (approved):

PR #765, opened 2022-08-19
Revert "fix(pid-182): Use correct input type for password component"

hectorakemp:
Reverts aesop/aesop-gel#764. Unfortunately making the input type a `password` type breaks the `show` button, which does not work on a password input box (and keeps it completely hidden). 
Maybe we could work around this by making the 'show' button toggle the input type between `text` when we want users to see the password, and `password`, when they don't? 

Open to suggestions.

Screenshots attached: 
image
image


cyrusxzw:
> Reverts #764. Unfortunately making the input type a `password` type breaks the `show` button, which does not work on a password input box (and keeps it completely hidden). Maybe we could work around this by making the 'show' button toggle the input type between `text` when we want users to see the password, and `password`, when they don't?
> 
> Open to suggestions.
> 
> Screenshots attached: image image

cyrusxzw (approved):

PR #767, opened 2022-08-19
fix(pid-182): Add simplified password input component with hide/show toggle

hectorakemp:
The reason that the ticket was raised was due to the password input field in the new login modal being misinterpreted as a regular `text` input. This caused password managers and Chrome android to misinterpret keystroke events. See video attached in the ticket for more information. 

This pull request creates a 'simple' password input component with a hide/show toggle. It is presented in the hide state as a stock `password` input, which allows password managers to interpret it correctly and autofill passwords. We intend to import it into the new login modal.

Instead of changing the original `PasswordInput` object, which contains lots of very subtle and well-done animation work, I've decided to duplicate the structure of that component, as well as its tests, types, and story, so that we have a choice given the situation on which component to use. 

To test: 
Verify that password managers can autofill the storybook component, and that the `onChange` function is printing the value of input field to the browser's console.


src/components/PasswordInputSimple/PasswordInputSimple.tsx
```diff
@@ -0,0 +1,76 @@
...
+        id={id}
+        isEnabled={isEnabled}
+        label={label}
+        name={name}
+        onChange={(e) => {
+          setValue(e.target.value);
+          onChange(e);
+        }}
+        theme={currentTheme}
+        type={isMaskingEnabled ? 'password' : 'text'}
```
cyrusxzw, line 57: Out of curious, why not just add `type={isMaskingEnabled ? 'password' : 'text'}` in `PasswordInput.tsx` rather than create a new component?

hectorakemp: That's a great point @cyrusxzw - however, it does break/make redundant the animation aspects of `PasswordInput`, as well as break the spec. I chose to make a new component instead, as we might want to use the animation work that was done in `PasswordInput` somewhere else that isn't the login modal.

hectorakemp: Originally, I thought we should just strip out all of the animation aspects of `PasswordInput`, and have it as a 'standard' password input object with hide/show capability. Do you think that's the best way to go for this change?

anthonycole: Is the password input used anywhere else @hectorakemp ? That's the only reason I'd keep this - to ensure backwards compatibility when merging it in. If it's not used anywhere else I'd just modify the existing component.

cyrusxzw: @hectorakemp , I'm not quite sure if animation is UI requirement. I didn't touch this component before. Any thought? @kevin-ho87 @JasonPingAesop

hectorakemp: Yes, it seems to be used in [this](https://github.com/aesop/aesop-web-ui/blob/develop/src/app/components-custom/checkout/register-user/registration-form.tsx#L31) component

hectorakemp: @anthonycole

cyrusxzw (approved):

PR #778, opened 2022-09-15
feat(ret-861): render CabinetIcon in MobileView if cabinetUrl provide…

ivandaho:
…d in GlobalNavigationContext

https://aesoponline.atlassian.net/browse/RET-861

Support `cabinetUrl` and `cabinetTitle` when using `GlobalNavigationContext` in [web-ui](https://github.com/aesop/aesop-web-ui/blob/24616a48edde69a6216a2adb5afcef1c9348cb6e/src/app/components-custom/navigation/Navigation.tsx#L250)

**Note: this PR is for mobile navbar, desktop will come later**

Current navbar consists of search, cart, and the menu. This PR adds an additional `<a>` to the cabinet/wishlist

Currently this only needs to be a static link, with no additional onclick handlers
We can conditionally provide the `cabinetUrl` prop when consuming in `web-ui` if Cabinet/wishlist is enabled/disabled

### Development Process
I followed the existing style, because it already has correct padding/spacing, and tab/focus CSS done

Note that this is different from the `AddToCabinetButton` in `aesop-engineering/wishlist`, even though they look very similar

Added `heart` to `Icon` with the svg provided from design team

I have put the UI code for the button in a separate file, with the code referenced from the current search icon. I don't think this is specifically needed to be exposed via a storybook story. I just felt it was better to have this in another file, and also eslint was complaining about the file length

Would appreciate input from Consideration team if this is ok, or if there is a better way to implement this

Screenshot 2022-09-15 at 3 29 49 PM


masonwongcs (approved):

src/components/GlobalNavigation/components/MobileView/components/Header/CabinetIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const CabinetIcon: React.VFC<CabinetIconProps> = ({
+  cabinetUrl,
+  iconClassSet,
+  title,
+}) => {
+  return (
+    <li className={styles.item}>
+      <a href={cabinetUrl}>
```
cyrusxzw, line 22: Why need both <a> and <button>? I think if it's only open the modal, it should be a button. If it navigate to a page, it should be a link.

ivandaho: Had some problems with accessibility on Firefox - couldn't tab to the links. But after searching a bit it's just a macOS problem. might have to enable this on macOS to tab to `<a>` elements. I pushed a commit to address this
Screenshot 2022-09-19 at 11 40 20 AM

cyrusxzw: wow, I have the same issue on firefox. I'll enable too. Thanks

